PR #1, opened 2018-06-11
Change color (blue is good but ...)

cmonjeau:
Change menu and primary panel color
Align admin panels

### DIFF
--- a/manager/views/users.html
+++ b/manager/views/users.html
@@ -75,7 +75,7 @@
             </div>
         </div>
     </div>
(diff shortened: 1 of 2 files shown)


manager/views/users.html
```diff
@@ -74,8 +75,10 @@ <h3 class="panel-title">Active</h3>
             </div>
         </div>
     </div>
+  <div>
```
osallou: should not be a "close" div  </div>?

PR #11, opened 2018-07-13
Fix mail renewal

mboudet:
Link for renewal does not work


osallou:
Missing index.html should not be an issue as it matches default page.

PR #18, opened 2018-10-02
Upgrade mongodb

cmonjeau:
Just change mongodb release to avoid security vulnerability on the repository

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,9 +16,24 @@
         "negotiator": "https://registry.npmjs.org/negotiator/-/negotiator-0.4.9.tgz"
       }
     },
-    "addressparser": {
-      "version": "https://registry.npmjs.org/addressparser/-/addressparser-0.3.2.tgz",
-      "integrity": "sha1-WYc/Nej89sc2HBAjkmHXbhU0i7I="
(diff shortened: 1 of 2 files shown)


osallou:
this only modifies python client , not mongodb itself (and won't upgrade, needs schema upgrades etc...)

mboudet:
👍

cmonjeau:
@osallou the security vulnerablitity concerns the 

> MongoDB bson JavaScript module versions 0.5.0 to 1.0.x before 1.0.5.

 Why update the module is not enough? Are you afraid that it will break some features? I'm confused and I want my T-shirt :)

osallou:
upgrading module only will not really impact security if server is not updated...

mboudet:
The vulnerability is not in mongodb, it is only in the plugin apparently.

https://www.cvedetails.com/cve/CVE-2018-13863/

If updating it does not break anything, we should probably do it. (Or at least correct it in prod)

osallou:
client should be tested against our prod version

cmonjeau:
I tested it with my dev version and it looks okay. I'll let you test it with your production version.

PR #19, opened 2018-10-02
Fix missing space in email

abretaud:
This patch will make the email much more readable


mboudet:
This is absolutely necessary. 👍

PR #22, opened 2018-10-03
Improve groups

cmonjeau:
Add user test for create and update group
Improve messages

### DIFF
--- a/routes/users.js
+++ b/routes/users.js
@@ -425,26 +425,33 @@ router.put('/group/:id', function(req, res){
     return;
   }
   users_db.findOne({_id: sess.gomngr}, function(err, user){
-    if(err || user == null){
-      res.status(404).send('User not found');
-      return;
-    }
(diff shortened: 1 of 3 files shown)


routes/users.js
```diff
@@ -425,26 +425,33 @@ router.put('/group/:id', function(req, res){
     return;
   }
   users_db.findOne({_id: sess.gomngr}, function(err, user){
-    if(err || user == null){
```
osallou: why removing this part?

routes/users.js
```diff
@@ -425,26 +425,33 @@ router.put('/group/:id', function(req, res){
...
-      res.status(404).send('User not found');
-      return;
-    }
     if(GENERAL_CONFIG.admin.indexOf(user.uid) < 0){
```
osallou: if user is null, will fail!!!

cmonjeau: This part is just below (line 432).

PR #33, opened 2019-04-16
Fix #32

### LINKED ISSUE #32: Format for projects expiration date
There is no help about the date format for project expiration.  
I tried dd/mm/yyyy and dd-mm-yyyy without any success. After that, using Month Day Year (e.g. Aug 31 2019) solved my case but it could be made more intuitive.
Why not give access to a calendar where the date can be set ?

mboudet:
Changed date format for projects to Day, Month Year

Added the calendar picker for project update


osallou:
when you modify the expiration date, where do you set it back to timestamp?

mboudet:
Yes
https://github.com/genouest/genouestaccountmanager/blob/master/manager2/src/app/admin/project/project.component.ts#L171

(It was already done)

PR #34, opened 2019-05-01
Fix projects

mboudet:


osallou:
@mboudet it is wrong to work on a day off ;-)

PR #38, opened 2019-06-11
[2] try to re-connect to mysql only if mysql is availlable

FanchTheSystem:


FanchTheSystem:
https://github.com/genouest/genouestaccountmanager/pull/44

PR #39, opened 2019-06-11
[3] Add some debug

FanchTheSystem:


app.js
```diff
@@ -11,17 +11,18 @@ var http = require('http');
...
 var winston = require('winston');
 var jwt = require('jsonwebtoken');
 
 const myconsole = new (winston.transports.Console)({
-      timestamp: true
+  label: 'gomngr',
+  level: log_level
 });
-winston.loggers.add('gomngr', {
+const wlogger = winston.loggers.add('gomngr', {
```
osallou: add function returns a logger itself?

could you check wlogger outputs were as expected?

FanchTheSystem: It look like it return this:
https://github.com/winstonjs/winston/blob/96591977288f7f1cd1bf38b7e8e629abb3e5f725/lib/winston/logger.js#L480
Do you want I add a check of the type in the code ?

FanchTheSystem:
https://github.com/genouest/genouestaccountmanager/pull/44

PR #40, opened 2019-06-11
[2] use lastest node version for dockerfile

FanchTheSystem:


.travis.yml
```diff
@@ -16,5 +16,5 @@ script:
   - npm install -g mocha
   - npm install -g nyc
   - cd $TRAVIS_BUILD_DIR/tests && docker-compose up -d
-  - cd $TRAVIS_BUILD_DIR && sleep 180 && docker ps && nyc --reporter=text mocha -- --timeout 10000
```
osallou: why removing the timout on tests ?

osallou: oh ok, I see it was moved to an option file

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8-jessie
+FROM node:latest
```
osallou: not fond of using latest.
May break at anytime when there is a  new release (not always backard compatible). Why not upgrading, but should specify a version.

FanchTheSystem: f0659c1afcc957a1cecb343e0adf7a5d47fd50cb

FanchTheSystem:
https://github.com/genouest/genouestaccountmanager/pull/46

FanchTheSystem:
Fixed: https://github.com/genouest/genouestaccountmanager/commit/2381009a0743b6f4560e5b5daab133ddc41dd901

PR #50, opened 2019-06-13
[4] Find ldap cn from uid 

FanchTheSystem:


FanchTheSystem:
Ready for review.
Some test with selenium here:
https://gitlab.cluster.france-bioinformatique.fr/taskforce/selenium-my/blob/master/Test-My-Create-User-Group-And-Change-Password.side

PR #51, opened 2019-06-13
[5] Add user home var

FanchTheSystem:


FanchTheSystem:
Some test with selenium here:
https://gitlab.cluster.france-bioinformatique.fr/taskforce/selenium-my/blob/master/Test-My-Change-Group-And-Check-Path.side

Still need to add a migration script before ask for review

FanchTheSystem:
https://github.com/genouest/genouestaccountmanager/pull/63

PR #57, opened 2019-07-11
[1] try to show an error if user id is empty on registration

FanchTheSystem:


osallou:
test already fixed in an other commit

PR #65, opened 2019-07-29
[2] Rename is_genouest to is_internal

FanchTheSystem:


mongo_rename_field.sh
```diff
@@ -0,0 +1,16 @@
+# replace field:
+# mongo gomngr --quiet --eval 'db.getCollection("users").updateMany( {}, { $rename: { "is_genouest": "is_internal" } } )'
+
+
+# add a new field with same value
+mongo gomngr --quiet --eval 'db.users.find().snapshot().forEach(
```
osallou: snapshot does not work for me (mongo 4.0.3)

2019-08-05T06:33:35.026+0000 E QUERY    [js] TypeError: db.users.find(...).snapshot is not a function :

osallou: I updated manually

FanchTheSystem: I think snapshot should be used as we update while we foreach

osallou: as we don't add/delete documents, should be ok not to use it.
anyway does not work on my mongo version, could not find why

PR #68, opened 2019-08-05
[4] allow any size for login logo

FanchTheSystem:


manager2/src/styles.css
```diff
@@ -206,8 +206,6 @@ button {
 
 .profile-img
 {
-    width: 74px;
```
osallou: maybe a max-width should be set to avoid ugly things or to manage larger images

FanchTheSystem: https://github.com/genouest/genouestaccountmanager/pull/68/commits/997a91b1d0f8341c9a42d3ed9f70f56d5d7a8ec8

PR #69, opened 2019-08-08
[5] allow to use custom path for express access log 

FanchTheSystem:
to allow split of log from app and express


osallou:
should not forget to add in README/install info about var MY_ACCESS_LOG

FanchTheSystem:
> should not forget to add in README/install info about var MY_ACCESS_LOG

https://github.com/genouest/genouestaccountmanager/pull/70

PR #72, opened 2019-08-26
Fix ssh key download link

FanchTheSystem:


manager2/src/app/user/user.component.html
```diff
@@ -402,9 +402,9 @@ <h4>SSH keys</h4>
                     <button class="btn btn-default" (click)="ssh_new_key()">Generate new key</button>
                     <hr/>
                     <h4 style="font-weight: normal;">Download</h4>
-                    <a [routerLink]="['/ssh', {id: user.uid}, 'private']" target="_blank"><button class="btn btn-default">private rsa key</button></a>
-                    <a [routerLink]="['/ssh', {id: user.uid}, 'putty']" target="_blank"><button class="btn btn-default">putty private key</button></a>
-                    <a [routerLink]="['/ssh/', {id: user.uid}, 'public']" target="_blank"><button class="btn btn-default">public key</button></a>
+                    <a [href]="[get_ssh_key(user.uid, 'private')]" target="_blank"><button class="btn btn-default">private rsa key</button></a>
+                    <a [href]="[get_ssh_key(user.uid, 'putty')]" target="_blank"><button class="btn btn-default">putty private key</button></a>
+                    <a [href]="[get_ssh_key(user.uid, 'public')]" target="_blank"><button class="btn btn-default">public key</button></a>
                 </div>
```
osallou: Did you test it?
Why redirecting to an other page instead of current routing? Redirecting to backend also prevent jwt usage.

osallou:
What is the issue here?
For info, admins cannot download their own keys.

manager2/src/app/user/user.component.html
FanchTheSystem: it is not a redirection, only the link string created from the service.

FanchTheSystem:
With current version I get link like this:
http://192.168.105.166:3000/manager2/ssh;id=myadmin/public
With the fix I always get the link where the api is installed:
http://192.168.105.165:3000/ssh/myadmin/public
Maybe it is due to angular version (RouterLink may not have same behavior), or something special in my configuration ...

manager2/src/app/user/user.component.html
osallou: A href is a rediection to backebd, not a route in frontend.
On href click, you are redirected to a new page. Routing is different, it loads the angular component.
It indeed seems there was no ssh route in manager2. I need to check what was done in manager 1. Seems something git wrong in port to new angular version.

FanchTheSystem: Oh ok :)  I can add a route in the frontend if needed, please keep me inform

osallou:
I will check code to see what was originally expected (even if does not work for the moment)

osallou:
I have refactored ssh download in commit 464264605ddb8de325e023b8d87b5f31d9105e40 , so I close this PR

PR #74, opened 2019-09-02
try to fix download key in firefox

FanchTheSystem:
Download key does not work in firefox on my computer

Fix based on:
https://stackoverflow.com/questions/30694453/blob-createobjecturl-download-not-working-in-firefox-but-works-when-debugging
https://stackoverflow.com/questions/32225904/programmatical-click-on-a-tag-not-working-in-firefox#

(build with **ng build --prod**)


osallou:
Did you test fix on chrome?

FanchTheSystem:
Yes I have tested in chrome (on ubuntu)

PR #77, opened 2019-09-05
Remove accent (and other non alphanum char) from userid

FanchTheSystem:
https://gitlab.com/ifb-elixirfr/ansible-roles/ansible-my/issues/10


lecorguille:
It's remove or replace?
é -> e
ö -> o

osallou:
remove...

lecorguille:
Hum, I will be sad if I lost my terminal e
Le Corguillé -> lecorguill

And I guess that it's worse but the accent is in the middle of the name :/

Gèrbe -> grbe

:P

FanchTheSystem:
@lecorguille : you can still add it by hand, if you insist I will propose another fix after my vacation ;)

PR #78, opened 2019-09-05
disallow user to register with an already used email

FanchTheSystem:


osallou:
This should be checked for fake users...
Fake users are interbal accounts and may have/share some emails (but not registered in mailing list)

PR #82, opened 2019-09-17
Show message from backend on user deletion

FanchTheSystem:


FanchTheSystem:
maybe it should be colored according to message

PR #85, opened 2019-09-19
Try to fix project list on user page

FanchTheSystem:


osallou:
Could you describe what you are trying to fix?
From what i see, i suppose you want to limit project list to projects user is not a member of when adding user to project.

FanchTheSystem:
I think the current list of project proposed on the user page is the same for **add** and **delete**. (you should check on your instance). I am not sure I understand why.

So I try a cleaner way with 2 list of project and it work on my instance. Now the list for **add** contain project which are not in the list for **del**

Also this pr fix a small typo for __this.add_to_project_msg__ variable

PR #91, opened 2019-09-23
Add and sort some list

FanchTheSystem:
For users, projects and groups always sort list in frontend.
+ Add users list in project creation and update (replace text input)

### DIFF
--- a/manager2/src/app/admin/project/project.component.ts
+++ b/manager2/src/app/admin/project/project.component.ts
@@ -19,7 +19,7 @@ export class ProjectComponent implements OnInit {
   project: any
   groups: any[]
   users: any[]
-  all_user: any[]
+  all_user: any
   prj_err_msg: string
   prj_msg: string
   oldGroup: string
@@ -49,7 +49,6 @@ export class ProjectComponent implements OnInit {
       path: ''
     }
     this.users = [];
-    this.all_users = [];
     this.groups = [];
   }
 
--- a/manager2/src/app/admin/project/project.component.ts
+++ b/manager2/src/app/admin/project/project.component.ts
@@ -19,6 +19,7 @@ export class ProjectComponent implements OnInit {
   project: any
   groups: any[]
   users: any[]
+  all_user: any[]
   prj_err_msg: string
   prj_msg: string
   oldGroup: string
@@ -48,6 +49,7 @@ export class ProjectComponent implements OnInit {
       path: ''
     }
     this.users = [];
+    this.all_users = [];
     this.groups = [];
   }
 
@@ -76,7 +78,11 @@ export class ProjectComponent implements OnInit {
     this.groupService.list().subscribe(
       resp => this.groups = resp,
       err => console.log('failed to get groups')
-    )
+    );
+    this.userService.list().subscribe(
+      resp => this.all_users = resp,
+      err => console.log('failed to get all users')
+    );
   }
 
   show_project_users(projectId) {
@@ -100,7 +106,7 @@ export class ProjectComponent implements OnInit {
       },
       err => console.log('failed to get project')
     )
-    
+
   }
 
   add_user(project, userId) {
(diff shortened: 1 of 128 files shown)


manager2/src/app/admin/project/project.component.ts
```diff
@@ -19,6 +19,7 @@ export class ProjectComponent implements OnInit {
   project: any
   groups: any[]
   users: any[]
+  all_users: any
```
osallou: any[] ? histoire de spécifier que c'est un tableau

manager2/src/app/admin/projects/projects.component.ts
```diff
@@ -38,13 +38,14 @@ export class ProjectsComponent implements OnInit {
 
   projects: any[]
   groups: any[]
+  all_users: any
```
osallou: same comment, any[] ?

manager2/src/app/user/user.component.ts
```diff
@@ -255,8 +239,6 @@ export class UserComponent implements OnInit {
             new_projects.push({id: projects[i].id, owner: false, group: projects[i].group, member: false});
         }
     }
-    user_projects.sort(this._compareId)
-    new_projects.sort(this._compareId)
```
osallou: why removal of sort on groups, user_projects and new_projects ?

manager2/src/app/user/user.service.ts
```diff
@@ -256,7 +258,13 @@ export class UserService {
     };
     return this.http.get(
       environment.apiUrl + '/user',
-      httpOptions)    
+      httpOptions).pipe(map((response: any) => {
+          return response.map(item => {
+            return item;
+          }).sort(function (a,b) {
```
osallou: why a response.map(item => return item) ?? If I am correct, this will return a copy of response.

manager2/src/app/admin/project/project.component.ts
FanchTheSystem: https://github.com/genouest/genouestaccountmanager/pull/91/commits/2627ae73646b35c3e5697c407d41de08f1144917

manager2/src/app/admin/projects/projects.component.ts
FanchTheSystem: https://github.com/genouest/genouestaccountmanager/pull/91/commits/2627ae73646b35c3e5697c407d41de08f1144917

manager2/src/app/user/user.component.ts
FanchTheSystem: I think they are useless as those list as created from already ordered list. I can re-add it if you want ;)

osallou: Are you sure they are ordered ?
Should not count anyway on order of API result (may change on backend).

manager2/src/app/user/user.service.ts
FanchTheSystem: Because it was done like this in groups services : https://github.com/genouest/genouestaccountmanager/blob/cf900751733845be443455ee457d9d366767bc38/manager2/src/app/admin/groups/groups.service.ts#L27

You are right, it is useless, and it look like it is cleaner to add some [] to first map and apply the sort on response: https://github.com/genouest/genouestaccountmanager/pull/91/commits/6a911bdd6e8bb5dd00bf134a2f3c46cc8d23276c

manager2/src/app/user/user.component.ts
FanchTheSystem: yes sort is done in the projects services in frontend:
https://github.com/genouest/genouestaccountmanager/pull/91/files#diff-fc06c10047d9960aa88b78aa835e45f8

FanchTheSystem: But as groups does not come from groups services (there are from user.group) I will re-add the sort for groups

FanchTheSystem: https://github.com/genouest/genouestaccountmanager/pull/91/commits/7743e72be3e6155f199fed854a7c21bbca76d067

PR #93, opened 2019-09-26
Indentation

FanchTheSystem:


FanchTheSystem:
`find . -name '*.js' -type f -not -path '.git*' -exec emacs -batch {} --eval '(setq indent-tabs-mode nil)' --eval '(indent-region (point-min) (point-max) nil)' -f save-buffer \;`

find . -name '*.ts' -type f -not -path '.git*' -exec emacs -batch {} --eval '(setq indent-tabs-mode nil)' --eval '(indent-region (point-min) (point-max) nil)' -f save-buffer \;

FanchTheSystem:
find . -name '*.js' -type f -not -path '.git*' -exec emacs  -batch -l ~/.emacs {} --eval '(setq indent-tabs-mode nil)' --eval '(indent-region (point-min) (point-max) nil)' -f save-buffer \;

PR #96, opened 2019-10-01
Set Window Title and Html Header Title to config.name + 'Account Manager'

FanchTheSystem:


manager2/src/app/app.component.html
```diff
@@ -3,7 +3,7 @@
   <nav class="navbar navbar-expand-lg navbar-fixed-top navbar-dark bg-dark navbar-inverse">
     <a class="navbar-brand" routerLink="/">
       <img class="d-inline-block align-top" style="width: 70px; height: 50px;" src="assets/images/logo-brand.png" alt="">
-          Account Manager
+          {{ config.name }} Account Manager
```
osallou: this does not create error?
config is not set at app setup time, so config.name is undefined.
Or config should be init with something like {'name': '' }

once config is loaded, this is fine.

FanchTheSystem: No error on my side, maybe because config is get in the "ngOnInit" event of app component
I added a default value as you ask:
https://github.com/genouest/genouestaccountmanager/pull/96/commits/d9766a7a12e6d13a6e34f01a9ea39b6745419fe6

FanchTheSystem:
Build is ok in travis since 19h ago, I don't know why github don't show it

PR #99, opened 2019-10-02
Check Login before send it to backend

FanchTheSystem:


manager2/src/app/auth/login/login.component.ts
```diff
@@ -90,32 +90,45 @@ export class LoginComponent implements OnInit {
...
 
     password_reset_request() {
-        this.msgstatus = 0;
-        this.msg = "";
-        if(this.userId == null || this.userId == "") {
-            this.msgstatus = 1;
-            this.msg = "Please enter your used id!";
-        }
-        else {
+        if (this._check_userId()) {
```
osallou: ?? you open the if here, but in diff I do not see curly bracket close

osallou: oh, okay, maps to the previous else...

PR #103, opened 2019-10-23
Add default template

FanchTheSystem:


osallou:
Rather than using symlinks woumd be better in file template manager to look at templates/xx/templatename and if not present use templates/default/templatename

Would avoid setting a symlink for all specific templates

FanchTheSystem:
https://github.com/genouest/genouestaccountmanager/pull/104

PR #104, opened 2019-10-23
Add default template (bis)

FanchTheSystem:
For this I need to do some test on my preprod, please wait before merge (review are welcome)


osallou:
Seems nunjucks already manage multi dir support, nice...

FanchTheSystem:
It look like it work well for ifb templates, so you should merge if you like it ;)

PR #108, opened 2019-10-29
Some ui fix and hide

FanchTheSystem:
https://gitlab.com/ifb-elixirfr/ansible-roles/ansible-my/issues/40
https://gitlab.com/ifb-elixirfr/ansible-roles/ansible-my/issues/3

### DIFF
--- a/manager2/src/app/auth/register/register.component.html
+++ b/manager2/src/app/auth/register/register.component.html
@@ -6,7 +6,7 @@
         </div>
         <form class="form-register form-group">
             <span class="glyphicon glyphicon-user" aria-hidden="true"></span>
-            <i>Personal informations</i>
+            <i>Personal information</i>
             <hr>
             <div class="form-group row">
                 <div class="col-md-4">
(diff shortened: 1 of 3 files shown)


osallou:
Don't understand the unsubscribe, if user is not active he is not subscribed anymore

FanchTheSystem:
> Don't understand the unsubscribe, if user is not active he is not subscribed anymore

It is to disable unsubscribe button for active user (as at ifb we don't want to allow people to unsubscribe from the cluster mailing list)

osallou:
config.newsletters etc... are hardcoded value for now, not taken from config file

manager2/src/app/auth/register/register.component.html
```diff
@@ -18,19 +18,19 @@
...
                 <div class="col-md-4">
-                    <label for="email">User id</label>
+                    <label for="email">Username</label>
                     <input autocomplete="username" [ngModelOptions]="{standalone: true}" [(ngModel)]="userid" type="text" id="userid" class="form-control" required maxlength="12"/>
 
                 </div>
             </div>
             <div class="form-group">
-                <label for="email">Professional email</label>
+                <label for="email">Professional academic email</label>
```
osallou: why academic ? We may have non academic

FanchTheSystem: you're right 74b9261e5120bfcd165953309a9d4ce535ef70ca

FanchTheSystem:
> config.newsletters etc... are hardcoded value for now, not taken from config file

sorry I don't understand, what I need to do (or not) ? ;)

PR #112, opened 2019-10-31
Auto add group on user activation

FanchTheSystem:


routes/users.js
```diff
@@ -994,9 +994,29 @@ router.get('/user/:id/activate', async function(req, res) {
 
     let data = await utils.mongo_groups().findOne({'name': user.group});
     if(!data) {
-        res.status(403).send('Group ' + user.group + ' does not exist, please create it first');
-        res.end();
-        return;
+        if (CONFIG.general.auto_add_group) {
+            let mingid = await utils.getGroupAvailableId();
```
osallou: does this behavior not already done somewhere else to call a function rather than "recoding" this (and avoid later issues) ?

The behavior is optional by config, but I would highly recommend not to do that (return of experience...).
Before, we used to do so, and lead to many errors in group names. Some people were in the same functional "group" but with different unix groups (  cnrsXX, cnrs_XX, umrXX, etc...) . Admin used to click "quickly" on activate without strict checking of group names.
This is usually only discover quite some time after, and we need to delete some groups, move users etc....

FanchTheSystem: > does this behavior not already done somewhere else to call a function rather than "recoding" this (and avoid later issues) ?
> 

Yes we can factorize with https://github.com/genouest/genouestaccountmanager/blob/35d61c084ecd9735560bcc05a26b13b90c3abb77/routes/users.js#L541

> The behavior is optional by config, but I would highly recommend not to do that (return of experience...).
> Before, we used to do so, and lead to many errors in group names. Some people were in the same functional "group" but with different unix groups ( cnrsXX, cnrs_XX, umrXX, etc...) . Admin used to click "quickly" on activate without strict checking of group names.
> This is usually only discover quite some time after, and we need to delete some groups, move users etc....

Yes I know, I have another PR in preparation to solve this issue, it will allow to use other value for default group (instead of team, use username or a specific group)

FanchTheSystem: Factorization done, I am testing on my dev env right now.

routes/users.js
```diff
@@ -131,12 +132,16 @@ var create_extra_group = async function(group_name, owner_name){
...
     try {
-        let created_file = await filer.user_create_extra_group(group, fid);
+        let created_file = await filer.user_add_group(group, fid);
         logger.info('File Created: ', created_file);
     } catch(error){
         logger.error('Create Group Failed for: ' + group.name, error);
         throw 'group creation failed';
     }
+
+    await utils.mongo_events().insertOne({'owner': owner, 'date': new Date().getTime(), 'action': 'create group ' + req.params.id , 'logs': [group.name + '.' + fid + '.update']});
```
osallou: owner and req.params.id are not known here

FanchTheSystem:
> if you plan an other PR to set username as group or other, should you not create a single PR for this as this will impact again the auto_add_group with a different behavior?

Here is Work In Progress PR:
https://github.com/genouest/genouestaccountmanager/pull/113
it don't impact the same method

PR #114, opened 2019-11-14
add a function to find group dn before ldap action on group

FanchTheSystem:


osallou:
What is the purpose here? Why getting group dn, for subgroups? *my* will not create subgroups.

routes/goldap.js
```diff
@@ -215,12 +274,25 @@ module.exports = {
...
+                .then(
+                    group_dn => { // resolve()
+                        return filer.ldap_add_user_to_group(user,group_dn, fid);
+                    })
+                .then(
+                    created_file => {
+                        logger.info('File Created: ', created_file);
+                    })
+                .catch(error => { // reject()
+                    logger.error('Delete Group Failed for: ' + group.name, error);
```
osallou: Wrong message, here we add user to group

FanchTheSystem:
> What is the purpose here? Why getting group dn, for subgroups? _my_ will not create subgroups.

It is to be able to add and remove user from ldap group not created in My
For example, in french: https://gitlab.com/ifb-elixirfr/ansible-roles/ansible-my/issues/56

routes/goldap.js
FanchTheSystem: a9bcd2bf89ae0097a98746aa28f8b2f75d330005

routes/goldap.js
```diff
@@ -215,12 +274,25 @@ module.exports = {
...
-        
+
         try {
             let created_file = await filer.ldap_add_user(user, group, fid);
             logger.debug('File created', created_file);
-            created_file = await filer.ldap_add_user_to_group(user, fid);
-            logger.debug('File created', created_file);
+
+            await get_group_dn(group.name)
+                .then(
```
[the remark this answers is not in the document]
FanchTheSystem: a9bcd2bf89ae0097a98746aa28f8b2f75d330005

osallou:
We should discuss about this.
From ldap, that's fine, but from local mongo, what uf group does not exists in my? (when adding/removing user to group)

There is no check on group in such cases?

osallou:
I did a quick check for case where group does not exists in db , looks fine as delete will silently fail

FanchTheSystem:
You can have imported group in mongo not created in my, for example if you plug my on an existing ldap and use import script.

Also If you change the ldif template for group creation, you can have any kind of dn you like for group or subgroup. 

So here we use the cn as identifier for the group to find the dn.

osallou:
Using name in ldap search for pre existing groups you may have duplicates.
In code you take first one, but may relate to an other one.
Should not  group in my rather define its dn?

FanchTheSystem:
Maybe we should add the dn as a mongo group field ... and Maybe it should be done for user too

osallou:
This could be done indeed, but should be optional to avoid complexity. If
dn is not defined, then use cn=x,ou=...
And would avoid to query for dn.

For users not sure if really needed as would logically be linked to main
group, and would complexify things if dn is modified.
For groups, would need to manage case where dn is modified.

osallou:
the pb with dn in interface, is to move a user/group to a different level, level must already exists.
So for imports, that's fine, but we should not allow to change dn, or we should find a way to create structure if it does not exists

example move

    dn: cn=group1,ou=groups,dc=xx

to

    dn: cn=group1,apps=ssh,ou=groups,dc=xx

FanchTheSystem:
In the last commit I add a default value if there is more than one result in the dn search.

so there is three case:
- people use default template and have an empty ldap
- people use default template and want to reuse existing group, so they are imported in mongo, and should be unique
- people customise template, so they have to manage unicity and level of group by themself (most of time they will customize template because they have other tool to manage ldap and or account creation)

I think all the 3 will work with current code

osallou:
will think about this a little more and test on my side. Expect to merge next week. If not done, please ping me

FanchTheSystem:
As I have some error in the build (in travis the group for admin is created to quickly before the user) and also because I want to use the auto_created_group feature :)

I have add a default value also if it don't find the dn.

Now it will search for dn, and if there is something not "normal" it return the default one.

So in fact the only case covered is when you use My on an existing ldap database which may have group with other level than the default for My

osallou:
What do you mean by  *group for admin is created to quickly* ?

routes/goldap.js
```diff
@@ -9,17 +9,82 @@ const filer = require('../routes/file.js');
...
+                logger.error('error ', err);
+                reject(err);
+                return;
+            });
+            // eslint-disable-next-line no-unused-vars
+            res.on('end', function(result) {
+                let default_dn = 'cn=' + group + ',ou=groups,' + CONFIG.ldap.dn;
+
+                if (group_dn_list.length > 1) {
+                    logger.error('more than one entry have been found', group_dn_list);
```
osallou: could you record an event with the error when this case if found , in addition to log.

FanchTheSystem:
> What do you mean by _group for admin is created to quickly_ ?

On empty database, like in travis or other ci tools, the admin and group user are created at the first start of the process. 

So you have a quick workflow like this => create group, create account, attach account to group.
When you are in the last step you can't know if the ldif file from the first step have been process, so you may not find any info about the dn in the ldap database

osallou:
In test env, script are executed immediatly, not managed via cron process,
so should be available

routes/goldap.js
FanchTheSystem: I think event are related (and attached) to user as owner, here we don't know the user, do you want an event with admin as owner ?

FanchTheSystem:
> In test env, script are executed immediatly, not managed via cron process, so should be available 

immediately in asynchronous world is relative ;)

osallou:
Should be via an await

osallou:
With admin

FanchTheSystem:
> With admin

194d0e707f7e4717af1db35f65d2cf8472efa692

routes/goldap.js
FanchTheSystem: 194d0e707f7e4717af1db35f65d2cf8472efa692

routes/goldap.js
```diff
@@ -9,17 +9,83 @@ const filer = require('../routes/file.js');
...
+                    logger.warn('switch to default value, you should check your ldap database and scripts logs results...');
+                    utils.mongo_events().insertOne({'owner': CONFIG.general.admin[0] , 'date': new Date().getTime(), 'action': 'find duplicate group dn for ' + group, 'status' : -1});
```
osallou: Please prefix message with an error tag: find duplicate group dn for
=> [error] found duplicate....

FanchTheSystem:
Please could you add a tag after merge (If you merge ;) )

routes/goldap.js
FanchTheSystem: 28a55a10cb1bcfe71b8d9bab859fcd7acb227ec0

PR #118, opened 2019-11-19
remove maxlength from html 

FanchTheSystem:
as there is a config param to allow any length since https://github.com/genouest/genouestaccountmanager/pull/109

### DIFF
--- a/manager2/src/app/auth/register/register.component.html
+++ b/manager2/src/app/auth/register/register.component.html
@@ -19,7 +19,7 @@
                 </div>
                 <div class="col-md-4">
                     <label for="email">Username</label>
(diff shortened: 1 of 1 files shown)


osallou:
but should not it be kept to force the limit (using config length)?

FanchTheSystem:
It will

> but should not it be kept to force the limit (using config length)?

It will need to add the length in the frontend conf service. It will add complexity for minor value ...

If the size is set to 256, it maxlenght  will be too big to be visible in the input.

Also as it may be an auto generated value from firstname and lastname,currently there is no check of lenght for those field, so the html maxlenght may be useless.

As you wish I can add it

PR #119, opened 2019-11-25
set username as readonly on registration form while there is no error…

FanchTheSystem:
… message

### DIFF
--- a/manager2/src/app/auth/register/register.component.html
+++ b/manager2/src/app/auth/register/register.component.html
@@ -19,7 +19,7 @@
                 </div>
                 <div class="col-md-4">
                     <label for="email">Username</label>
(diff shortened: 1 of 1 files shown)


FanchTheSystem:
https://gitlab.com/ifb-elixirfr/ansible-roles/ansible-my/issues/11

osallou:
setting as readonly if there is no error may not really fit.
If error is related to empty address, or whatever, it will switch to non-readonly.
I do not understand by the way why user should not select his user id (as long as it doesnt already exists).

We provide a way to prefill it with first/lastname, as a convenience and ease logins in general, but user may select personal identifier to get the same in several environements, or easier logins depending on their name....

FanchTheSystem:
I don't understand why too ;) 

@lecorguille

PR #124, opened 2019-12-04
use toLocaleString and toLocaleDateString

FanchTheSystem:


FanchTheSystem:
Maybe with should use toLocaleDateString everywhere as we don't care about minute in My date

osallou:
What is output format?
I think i used specific function to format date because default was display full day instead of numbers, and longer strings rather than 'easy' date format like yy/mm/dd

FanchTheSystem:
> What is output format?
> I think i used specific function to format date because default was display full day instead of numbers, and longer strings rather than 'easy' date format like yy/mm/dd

As it use local, it will depend on user locale config (in browser)
In my config it show: 03/12/2020 with the time or not (depending if we use toLocaleDateString or to LocaleString)

FanchTheSystem:
```
const event = new Date(Date.UTC(2012, 11, 20, 3, 0, 0));
console.log(event.toLocaleDateString());
const options = { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' };
console.log(event.toLocaleDateString('fr-FR', options));
```

Give me

> "20/12/2012"
> "jeudi 20 décembre 2012"

PR #127, opened 2019-12-05
Add default home in config

FanchTheSystem:


FanchTheSystem:
It is because at ifb we want to have a project approch to cluster access.

For example you can't run slurm job, if you don't have a project. 

It is also for this that I added script (and template) for project creation and user attach.

I will add a pr (I think today) to add a form in project to ask(by mail) admin for a new project.

If you have any idea to improve the project approch to cluster access in my, please feel free to say it to me.

osallou:
adding a form, why not, but on our side user sends a helpdesk ticket (form which will end in helpdesk anyway on our side)

PR #138, opened 2019-12-12
show api key form only if user is admin

FanchTheSystem:


FanchTheSystem:
image

FanchTheSystem:
It don't know if user need to have access to this, and it may be confusing ....

osallou:
API Key is used to query API. Though there is little chance that our users really make use of this, this *may* be used and is a user personal info, so it should be displayed.

FanchTheSystem:
I can't find any use case of user needing api key. Do you have an example of a not admin use it ?

osallou:
user *may* use API to get personal data access, create database, upload ssh key, get quota info etc....

There is indeed low chance to be used by users, but could allow integration with other tools, user only providing is API key.
Could be move however to an other tab rather than on main tab.

FanchTheSystem:
ok

PR #140, opened 2019-12-13
fix template for user group in ifb cluster

FanchTheSystem:


FanchTheSystem:
Please could you tag after merge

templates/ifb/user/add_user.sh
```diff
@@ -20,7 +20,7 @@ touch "{{ user.home }}/.ssh/authorized_keys"
 echo 'Host *' > "{{ user.home }}/.ssh/config"
 echo '  StrictHostKeyChecking no' >> "{{ user.home }}/.ssh/config"
 echo '  UserKnownHostsFile=/dev/null' >> "{{ user.home }}/.ssh/config"
-chown -R {{ user.uidnumber }}:{{ user.gidnumber }} "{{ user.home }}"
+chown -R {{ user.uid }}:{{ user.uid }} "{{ user.home }}"
```
osallou: I usually prefer and advise to use ids rather than names.
If scripts were to be executed in a container or different server not connected to ldap, using name fail while id always work.

FanchTheSystem: yes I know we already have this conversation ;)

FanchTheSystem:
I have some gidnumber not well synchroniszed between my and ifb ldap. 

Maybe I will create a PR next week about this issue (minuid and mingid value) but I have to investigate more before this.

please consider it as a temporary fix which impact only ifb

FanchTheSystem:
please note that on ifb cluster, "chown -R {{ user.uid }}:{{ user.uid }} "{{ user.home }}"" will always work

PR #144, opened 2019-12-19
Disable user group

FanchTheSystem:
It add a variable which allow to disable user group (user.group)
so with this option, you will only have secondary group availlable.

At IFB we have another tool  to manage Ldap account, it is Num and it called in some template, here are some example:

- https://github.com/genouest/genouestaccountmanager/blob/master/templates/ifb/user/delete_user.sh
- https://github.com/genouest/genouestaccountmanager/blob/master/templates/ifb/project/add_user_to_project.sh

It may be used outstide of My, and it have to manage ldap user and project group and configure acl and slurm according to our need.

Recently we have had some issue due to gidnumber synchronisation for chown (and some other)

I think I will add some feature to Num to allow more call from My template in future PR (like for example to update user)

Also I think I will create another PR to check if gidnumber is used in ldap before creating a group in My.

So to resume this PR goal is to add a variable in config to allow to disable the user group management in My.
When I say user group, it is only "the" user main group (which is linked to user entry in ldap). The goal is to keep using secondary group for service or application group management.

Maybe I will add the same thing for project group (or maybe i will do it in another PR)

Also If we will use this option at IFB we don't need anymore the option I added in this PR:
https://github.com/genouest/genouestaccountmanager/pull/112
So Maybe I can remove it.

Will there is WIP in this title PR, it mean that I am not ready for merge, but you can start comment or discuss if you want ;)


osallou:
will certainly have many impacts in code and templates to remove group management.
though with use of async/await, will be easier to manage conditions in code....

check on gidnumber, from *my* point fo view should not be necessary as we allocate it from min grou id, always increasing. If you set the minimum group id to the higher value of your existing ldap, there should be no conflict.

FanchTheSystem:
> will certainly have many impacts in code and templates to remove group management.
> though with use of async/await, will be easier to manage conditions in code....

I do not know if we need to update default template, as there are "default" template, maybe they should work only with "default" config value... (I think most of the current config value are already not managed in current default template)

The impact in code is not so huge, as it will not remove it from mongo, it will juste use '' as user.group, so the only impacted code, is the one which manage this value

> 
> check on gidnumber, from _my_ point fo view should not be necessary as we allocate it from min grou id, always increasing. If you set the minimum group id to the higher value of your existing ldap, there should be no conflict.

only if there is no insert in ldap database from elsewhere, but is is not the main purpose of this pr, maybe we should try to keep discussion on user.group topic ;)

FanchTheSystem:
I think it work (at least with my test)
Please could you review it ?

FanchTheSystem:
I have not yet check for tp reservation

osallou:
Will check but i will also need t test tp reservation, though should not have an impact for us

routes/goldap.js
```diff
@@ -265,19 +265,25 @@ module.exports = {
...
-
         try {
             let created_file = await filer.ldap_add_user(user, group, fid);
             logger.debug('File created', created_file);
-            let group_dn = await get_group_dn(group.name);
-            created_file = await filer.ldap_add_user_to_group(user,group_dn, fid);
-            logger.debug('File created', created_file);
+
+            // should we do if (group) ?
+            if (!CONFIG.general.disable_user_group) {
```
osallou: Can't we just test group not null to avoid config many times.

osallou:
For tp reservation, a group is created (all tp temporary users are in same temporary group).
Group id is taken from getGroupAvailableId().

Group is deleted with users at the end of the reservation. Won't it have an impact on your side if feature is activated?

FanchTheSystem:
> For tp reservation, a group is created (all tp temporary users are in same temporary group).
> Group id is taken from getGroupAvailableId().
> 
> Group is deleted with users at the end of the reservation. Won't it have an impact on your side if feature is activated?

Yes I have to check

routes/goldap.js
FanchTheSystem: https://github.com/genouest/genouestaccountmanager/pull/144/commits/74c65173ac96d3cbfafdce8477501dfb930a403a

FanchTheSystem:
> Config has quite many option now, would be nice to explain option in config template

Yes I think I will create another PR with a config.md doc which will try to explain all the config variable

FanchTheSystem:
> Will check but i will also need t test tp reservation, though should not have an impact for us

As group is important for tp, if the config is set too disable I add it as secondary group
38270b173c642c5ab45acbb799824f7e88fcef86

I will do a test on my dev server tomorow

FanchTheSystem:
So I think this is PR is ok for me, please merge or comment ;)

routes/tp.js
```diff
@@ -89,8 +89,8 @@ var create_tp_users_db = function (owner, quantity, duration, end_date, userGrou
                 address: '',
                 lab: '',
                 responsible: owner,
-                group: userGroup.name,
-                secondarygroups: [],
+                group: (CONFIG.general.disable_user_group) ? '' : userGroup.name,
+                secondarygroups: (CONFIG.general.disable_user_group) ? [userGroup.name] : [],
```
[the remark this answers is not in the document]
FanchTheSystem: group creation is done before user creation, see router.exec_tp_reservation

PR #148, opened 2020-01-07
Add duration in config

FanchTheSystem:


manager2/src/app/auth/register/register.component.html
```diff
@@ -70,11 +70,7 @@
             <div class="form-group">
                 <label for="duration">Duration</label>
                 <select id="duration" [ngModelOptions]="{standalone: true}" [(ngModel)]="duration" class="form-control" required>
-                    <option value="90">3 months</option>
-                    <option value="180">6 months</option>
-                    <option value="365">1 year</option>
-                    <option value="730">2 years</option>
-                    <option value="1095">3 years</option>
+                  <option *ngFor="let duration of config.duration">{{duration}}</option>
```
osallou: should set value of option, will display in "month/year" but api expects days

FanchTheSystem: I don't think is a good idéa to have the day in frontend (anyone could set it to 999999999999 ....) 

So I have modified duration to be a string, so it contain only user input, and expiration is calculated using this input to find the number of day

FanchTheSystem: Do you need a script to update duration in mongo ?

osallou: I don't know if it is necessary, duration is only at registration time only I think.
Or is it reused when extending/reactivating user (admin or self-extend) ? Which duration is used at this time ?

osallou: Arrg, extend is using user duration field....

            let expiration = new Date().getTime() + 1000*3600*24*user.duration;

so it needs to be updated in db..... 
The problem to update in db is if there are some custom durations....

FanchTheSystem: I will add a script which find the nearest duration if it is custom value.
(I will do it tomorow)

FanchTheSystem: here is the script: https://github.com/genouest/genouestaccountmanager/pull/148/commits/cab37ade5909d2e865a055ce5facfe9bb558d193

I have tested it with some random generate number

FanchTheSystem:
So I think this is PR is ok for me, please merge or comment ;)

osallou:
you have a conflict on conf.js

FanchTheSystem:
conflict resolved, please could you tag after this PR ?

osallou:
tagged 1.3.0

FanchTheSystem:
> tagged 1.3.0

Thank you ;)

PR #150, opened 2020-01-09
[WIP] check if there is redis conf before modify uid

FanchTheSystem:
https://github.com/genouest/genouestaccountmanager/pull/149

### DIFF
--- a/routes/tp.js
+++ b/routes/tp.js
@@ -125,12 +125,14 @@ var create_tp_user_db = async function (tp_user) {
     let user = {...tp_user};
     logger.debug('create_tp_user_db', user.uid);
     try {
-        let uid = await utils.getUserAvailableId();
-        user.uid = CONFIG.tp.prefix + uid;
-        user.lastname = uid;
-        user.email = CONFIG.tp.prefix + uid + '@fake.' + CONFIG.tp.fake_mail_domain;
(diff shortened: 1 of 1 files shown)


routes/tp.js
```diff
@@ -125,12 +125,14 @@ var create_tp_user_db = async function (tp_user) {
     let user = {...tp_user};
     logger.debug('create_tp_user_db', user.uid);
     try {
-        let uid = await utils.getUserAvailableId();
-        user.uid = CONFIG.tp.prefix + uid;
-        user.lastname = uid;
-        user.email = CONFIG.tp.prefix + uid + '@fake.' + CONFIG.tp.fake_mail_domain;
-        user.uidnumber = uid;
-        user.home = fusers.user_home(user);
+        if (CONFIG.redis !== undefined && CONFIG.redis.host !== undefined && CONFIG.redis.host !== null) {
```
osallou: I'd like to keep separate id allocation management from other modules.
tp and others should not care if we use redis or any other management function, it just want an id...

FanchTheSystem: yes this pr does not intend to be merged, it is only to illustrate https://github.com/genouest/genouestaccountmanager/pull/149#issuecomment-572585965

PR #152, opened 2020-01-14
try to clean conf init value

FanchTheSystem:


routes/conf.js
```diff
@@ -40,57 +60,30 @@ function get_conf () {
...
-
-    let duration = Object.keys(get_conf().duration);
-
     let config = {
-        'main_groups': CONFIG.general.main_groups,
-        'terms_of_use': terms_of_use,
-        'default_home': default_home,
-        'name': CONFIG.general.name,
-        'support': CONFIG.general.support,
+        'main_groups': get_conf().general.main_groups,
```
osallou: better to call get_conf once before, then use object
during object dict, multiple get_conf will check for is_init, which may not be set yet.

FanchTheSystem: https://github.com/genouest/genouestaccountmanager/pull/152/commits/c7ef5964a0b32052bbeac678d36bf63bee77117a

PR #153, opened 2020-01-15
[WIP] add a new document about config variable

FanchTheSystem:


osallou:
Could use title for sections in config then lists for params

PR #165, opened 2020-03-26
remove some method from utils.js as the feature are handle by templates

FanchTheSystem:
there is no call of those method elswere in the code


FanchTheSystem:
travis is dead ?

osallou:
I thnk you (or someone before) moved those functions directly in plugin scripts, correct?

FanchTheSystem:
> I thnk you (or someone before) moved those functions directly in plugin scripts, correct?

not in plugin, in template

as example, see include here : https://github.com/genouest/genouestaccountmanager/blob/b44d0b462293134e5c34fd56087d0e66c7b77e82/templates/default/user/add_user.sh#L14

of this : https://github.com/genouest/genouestaccountmanager/blob/b44d0b462293134e5c34fd56087d0e66c7b77e82/templates/default/user/add_user.sh#L14

PR #171, opened 2020-04-29
fix value undefined in some mail

FanchTheSystem:
some project have some value undefined in ifb server
so mail failed on it
small fix

oups !

### DIFF
--- a/routes/utils.js
+++ b/routes/utils.js
@@ -477,7 +477,7 @@ async function gen_mail_opt (options, variables)
     // replace variable in message
     for (let key in variables) {
         let value = variables[key];
-        if (!value) { value = '';} // value may not be defined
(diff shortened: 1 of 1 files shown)


routes/utils.js
```diff
@@ -477,9 +477,11 @@ async function gen_mail_opt (options, variables)
     // replace variable in message
     for (let key in variables) {
         let value = variables[key];
+        if (!value) { value = '';} // value may not be defined
```
osallou, line 480: Should test undefined/null rather that ! for possible (future) boolean values

FanchTheSystem: https://github.com/genouest/genouestaccountmanager/pull/171/commits/66f27bd7ae825cd057f12776748318eee7f328c2

PR #173, opened 2020-05-04
fix user link for admin received email on user registration

FanchTheSystem:


osallou:
same kind of link is set in several places in user.js

for example

    let link = GENERAL_CONFIG.url + encodeURI('/user/'+uid+'/confirm?regkey='+regkey);

but should it be really sent to /manager2? Is is not expected to be redirected to backend which will then redirect to web ui? (need to check)
or should other links be updated too? (used to work at least)

osallou:
ok, this one looks specific, just specifying for user web page on my in mail. Is  this correct?

FanchTheSystem:
yes this link is specific, it should not be addressed to backend
It use for admin to have a direct link to user info in my, so he can check it and validate account in the ihm

PR #177, opened 2020-05-12
set permission on extradir every time they are moved

FanchTheSystem:


templates/default/user/move_extra_dirs.sh
```diff
@@ -15,9 +15,10 @@ then
     else
         mkdir -p "$extra_dir"
     fi
-    chown -R {{ user.uidnumber }}:{{ user.gidnumber }} "$extra_dir"
 fi
 
+chown -R {{ user.uidnumber }}:{{ user.gidnumber }} "$extra_dir"
```
osallou, line 20: one issue is we gonna change ownership even if not necessary, which depending on content can be quite long.
should change only if gid is different

FanchTheSystem: https://github.com/genouest/genouestaccountmanager/pull/177/commits/f6096e2c209398cd4de0562b2c1313c9eacbe665

templates/default/user/modify_user.sh
```diff
@@ -16,11 +16,23 @@ if [ "$oh" != "$nh" ]
 then
     mkdir -p $(dirname "$nh")
     mv "$oh" "$nh"
-    chown -R {{ user.uidnumber }}:{{ user.gidnumber }} "$nh"
+
+    {% if user.oldgidnumber %}
```
osallou, line 20: this test is I think useless, for user home, because if home moved, it means that his main group has been modified.
but anyway, it only a test which should always be true, so should be fine

FanchTheSystem: Yes it is useless check for now, maybe we can keep anyway, as :
- It does not change the current shell script generated 
- I like the idea that everytime when chown is used in the context of user modification, the check is done before run it

osallou: yeap

PR #189, opened 2020-10-07
Clean root file

FanchTheSystem:


Dockerfile
```diff
@@ -29,6 +29,7 @@ COPY templates /root/genouestaccountmanager/templates
 COPY views /root/genouestaccountmanager/views
 COPY tests /root/genouestaccountmanager/tests
 COPY test /root/genouestaccountmanager/test
+COPY cron /root/genouestaccountmanager/cron
```
osallou, line 32: not useful, cron is used for my-cron which Dockerfile is in cron/ so *my* container does not need cron directory

osallou: Image is based on my image but needed cron files are copied in cron image

PR #190, opened 2020-10-07
remove unused directory and file

FanchTheSystem:


osallou:
there is a build error due to remaining stuff in Dockerfile

Step 23/31 : COPY views /root/genouestaccountmanager/views
COPY failed: stat /var/lib/docker/tmp/docker-builder416470517/views: no such file or directory

PR #205, opened 2020-11-23
remove search and paginator on user pending table

FanchTheSystem:
because it is ugly and useless


osallou:
why? we may have quite many pending users, search and filter would ease work for admin

FanchTheSystem:
When there is no user, it is very confusing to have 4 search box in the same page.  And when there is some user, it is not very helpfull, if you keep a small number of pending user.

At IFB we never have more than 10 pending user, and most of time is 0. 

If you use it at genouest, we can keep those search box, it is ok for me ;)

PR #221, opened 2021-01-28
change projects page order

FanchTheSystem:
to have the same "box" order as in groups or users pages
creation on top, and list at the end

### DIFF
--- a/manager2/src/app/admin/projects/projects.component.html
+++ b/manager2/src/app/admin/projects/projects.component.html
@@ -1,236 +1,245 @@
-  <div *ngIf="notification" class="alert alert-success">{{notification}}</div>
-  <div class="card bg-light">
-    <div class="card-header" (click)="requests_visible=!requests_visible" style="cursor: pointer;">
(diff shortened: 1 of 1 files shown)


FanchTheSystem:
Also I think the accept button should be removed. 
As it is useless, it do the same as create, and it may confuse people.

osallou:
Need to check, but with konogan updates i think the accept button prefills the new project fields. @KonoganB ?

FanchTheSystem:
Yes there are 2 button, both pre-fills the fields and the "create" one do also one more task, it call add_project

here is the code:
https://github.com/genouest/genouestaccountmanager/blob/master/manager2/src/app/admin/projects/projects.component.ts#L301-L309

osallou:
ok so you mean to remove accept for automatic accept, and keep only modify to prefill new project, then admin creates it, or not, right?

FanchTheSystem:
> ok so you mean to remove accept for automatic accept, and keep only modify to prefill new project, then admin creates it, or not, right?

Yes it is, and it can be done in another PR (as this one is only for box order)

PR #224, opened 2021-01-28
limit the size of the name when asking for project

FanchTheSystem:


osallou:
I am fine with this, but you faced such issur?

PR #229, opened 2021-02-02
Move code

FanchTheSystem:
Same as here https://github.com/genouest/genouestaccountmanager/pull/228 + Some code move


osallou:
Please do not forget to update base scripts (tp reservation, etc...)

FanchTheSystem:
I think it work, and it can be a good start.

Please Review ;)

PR #232, opened 2021-02-16
fix group creation for admin when ldap is empty

FanchTheSystem:


osallou:
? I don't understand the "fix", it just calls a method copied elsewhere

FanchTheSystem:
> ? I don't understand the "fix", it just calls a method copied elsewhere

Yes it don't find the "elsewhere" by itself so I add grpsrv to say it is in the groups.service import

PR #237, opened 2021-03-03
Add isadmin

FanchTheSystem:
Add a service to check right and update code to use it instead of checking config each time.

It should be usefull to use other option to list admin (like a ldap group or any future way we want to implement)

And also it is start of service for other role in future feature like for exemple financial manager account


core/right.service.js
```diff
@@ -0,0 +1,18 @@
...
+//const winston = require('winston');
+//const logger = winston.loggers.get('gomngr');
+const CONFIG = require('config')
+
+// This will allow to change the way admin right are checked without impact on other part of the code
+// It will also be used to manage new role in future feature
+
+exports.is_admin = is_admin;
+
+async function is_admin(uid) {
```
osallou, line 11: Why async? Not needed here..   for future async ops?

core/right.service.js
```diff
@@ -0,0 +1,18 @@
+//const Promise = require('promise');
```
osallou, line 1: Rights as module name is not a good name, related to usage should rather be something like roles

osallou (commented):
External is admin is nice for future changes, could define admins to also be in a group, would be more dynamic (while keeping confug base for compat and easier startup).
And indeed further roles could be added in future though will be more complex on ui side..  as said in reviews, please change to *roles* or alike for module name (or acls).

core/right.service.js
FanchTheSystem: Yes I think about put the admin list in mongo database or maybe in an ldap group (maybe as an option or several option)

core/right.service.js
FanchTheSystem: yes I will rename it

routes/ssh.js
```diff
@@ -45,13 +43,25 @@ router.get('/ssh/:id/private', async function(req, res) {
...
+        return;
+    }
+
     if(!user) {
         res.send({message: 'User does not exist'});
         res.end();
         return;
     }
-    if(CONFIG.general.admin.indexOf(user.uid) >= 0){
+    // todo maybe remove this a next if do the job, and it will allow admin to download it's own private key
```
osallou, line 63: Nope, pure security stuff to avoif getting admin key

PR #240, opened 2021-03-08
replace var by const for every require

FanchTheSystem:


osallou:
Why replace by var?

FanchTheSystem:
> Why replace by const?

For the same require in two distinct file it is sometime "var" and sometime "const", so the first goal is to have only one syntax for all.

And I think const is better for require as those variable will never be modified. But if you want we can use var every where, I am ok with that too.

Issue #243, opened 2021-03-16
Add ability for user to leave a project


pierrepo:
Would it be possible to have the possibility for a user to leave a project (he does not own)?

osallou:
project owner can remove a user, but indeed, a user should be able to do it himself

osallou:
fixed in next release

PR #244, opened 2021-03-16
let user leave a project, links #243

osallou:
to be tested


osallou:
added removal of user from project group if any

PR #245, opened 2021-03-18
prevent reusing a user uid

osallou:
When a user is deleted, account is deleted. If later on someone
register with an old uid, depending on available apps, user may
gain access to old user jobs, data, etc.
To prevent this we record in old_users the uid of user when deleted
and at register with check that uid was not used.

Impact is if a legitimate user register again, he will not be able
to reuse his uid. However, before account validation, an admin
can modify his uid to reuse old one.


FanchTheSystem:
Here are my maybe list about it:

- Maybe we oldusers collections should be purged at a periodic date 

- Maybe we should keep it in the users collection and add a flag to say it is deleted 

- Maybe we should add a button in my to purge data (button availlable only on deleted profile) 

- Maybe the check should be on the uidnumber and gidnumber, as it is those value on most file system/acl which will give access to data.

- Maybe it should also be done for group and project as I think there are the same kind of issue (create a new group with the name of an old one)

- Maybe user data should be deleted when user is deleted.

osallou:
* Maybe we oldusers collections should be purged at a periodic date

If we purge, then uid can be reused at some time... this is waht we want to avoid

* Maybe we should keep it in the users collection and add a flag to say it is deleted

possible, but impacts API to filter queries on users AND for rgpd, we will also need at some time to delete all other user info

* Maybe we should add a button in my to purge data (button availlable only on deleted profile)

if we implement purge, yes, but as I said, this is not the goal

* Maybe the check should be on the uidnumber and gidnumber, as it is those value on most file system/acl which will give access to data.

[u|g]idnumber are incremental, so cannot be reused. Furthermore, as user data are deleted, reusing uid/gid is not an issue.

Reusing user id is not really an issue on linux side as long as data are deleted with account. The issue is more with uid with external apps using ldap. Usually, an app binds to the ldap but creates a local account (but still use ldap for auth).
When account is deleted on *my* side, it is removed from ldap, but not from app. If someone register using an old uid and access the app, he will gain access to data of old user.
In perfect world, at user deletion on *my* side, all user references should be deleted on all apps, but it is sometimes difficult....

* Maybe it should also be done for group and project as I think there are the same kind of issue (create a new group with the name of an old one)

For projects, it is only a functional name, so should have no impact on re-using it.
For groups, user does not select a group name, only admin, so should not really occurs. But ldap group name in external apps could indeed have same side effects, would make sense to control the same way.

* Maybe user data should be deleted when user is deleted.
When user is deleted, data (home dir and related dirs) are deleted, at least on our side. Doing the same on all external app is more complex. We do this via plugins for *some* apps, but not all.....

FanchTheSystem:
Please note that at IFB we use project name to create a slurm account and we also use it to give acl access to project folder. So in our case it is not only a "functional" name.

About user data and external app, you will have to find a way to delete them to respect the RGPD too. (Disallow other potential new user to access it is not a solution)

About the other point, I am OK with you ;)

It will break my selenium test suite as it re-use the same username on each test.  
And I think there are some other case I missed, where it will be usefull to be able to disable it.

So maybe we should add a config flag to enable or not the check on olduser collection.

osallou:
For rgpd, we have means to do so, but rgpd does not require things to be automatic...
Anyway, via plugins we have ways to do so, but requires more work and time...

Preventing reuse is a protection.

For projects in your case slurm account account is deleted with user. Recreating account in slurm will have no side effect (unless slurm keeps some data even on deletion).

For tests, we could set old (user/group) controls optional, with default set to true, if disabling control for tests (or willing to do so)

FanchTheSystem:
:+1:

osallou:
If we want full strong rgpd, we could store md5 of user uid instead of uid
itself..

osallou:
* added md5 of userid for old users
* enable/disable old users/groups/... by conf, enabled by default

PR #249, opened 2021-03-23
move 'remove user from' to core

FanchTheSystem:


FanchTheSystem:
@osallou  I tested it, it look ok, please could you review ?

PR #250, opened 2021-03-24
add a check on prevent_reuse config option before fail on insert

FanchTheSystem:
as this option is usefull on dev env for test

and in this kind of env you may want to enable or disable it sometime

so you may find data in the old table

so this merge request add a check on this var in the insertion check too

### DIFF
--- a/routes/users.js
+++ b/routes/users.js
@@ -476,7 +476,7 @@ router.post('/group/:id', async function(req, res){
         return;
     }
     let groupexisted = await dbsrv.mongo_oldgroups().findOne({name: req.params.id});
-    if(groupexisted){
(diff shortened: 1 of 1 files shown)


osallou:
why not, but not really necessary
check is done to insert data in **old**. If option is not set, nothing is put in **old**

so at check time, will always look in an empty table.

your check avoids to look in db if not needed

FanchTheSystem:
Yes I have seen, the current check does not "prevent reuse" (as the variable name say) of the username, it "prevent insert in old db" ;)

I think we have added this config option for test scenari,  and in test you may want to enable or disable this check.

And you may need to enable it and don't want to drop the old collection before your run your test. 

For example when you run a test with a prod like dumped mongo database.

PR #253, opened 2021-03-30
Add config service

FanchTheSystem:

PR #254, opened 2021-03-30
Add groups route

FanchTheSystem:


FanchTheSystem:
I will rebase when https://github.com/genouest/genouestaccountmanager/pull/253 will be merged

PR #260, opened 2021-04-08
fix init of user project array check

FanchTheSystem:


osallou:
This is the single place where undefined projects could be an issue?

PR #266, opened 2021-04-15
start using run_plugin

osallou:
Use new function *run_plugins* in plugin core service.

Instead of plugin_call and Promise.all multiple calls, just using something like:

    try {
        await plgsrv.run_plugins('activate', user.uid, user, 'auto');
    } catch(err) {
        logger.error('activation errors', user, err);
    }

run_plugins takes first argument as plugin function to call. If not present, will log and ignore the plugin for this call.

all plugins are executed sequentially with await, if one plugin fails, other plugins are executed anyway

it returns a boolean indicating if at least one error occured

### DIFF
--- a/core/user.service.js
+++ b/core/user.service.js
@@ -1,4 +1,3 @@
-const Promise = require('promise');
 const winston = require('winston');
 const logger = winston.loggers.get('gomngr');
 const crypto = require('crypto');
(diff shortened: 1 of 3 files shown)


osallou:
@FanchTheSystem , this makes plugin calls cleaner in code, unit tests are ok

FanchTheSystem:
It is more KISS :) so I am OK :)

PR #277, opened 2021-07-05
add a message to check spam on registration

FanchTheSystem:
maybe manager2/src/app/callback/registered/registered.component.html is not used by the code (I do not check)

### DIFF
--- a/manager2/src/app/callback/registered/registered.component.html
+++ b/manager2/src/app/callback/registered/registered.component.html
@@ -5,7 +5,9 @@
         </div>
         <div class="panel-body">
             <p>Your account creation has been taken into account.</p>
+            <br>
             <p>You will receive an email soon to <b>confirm your email address</b>.</p>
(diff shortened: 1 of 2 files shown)


manager2/src/app/callback/registered/registered.component.html
```diff
@@ -5,9 +5,11 @@ <h2 class="panel-title">Congratulation</h2>
         </div>
         <div class="panel-body">
             <p>Your account creation has been taken into account.</p>
+            <br>
             <p>You will receive an email soon to <b>confirm your email address</b>.</p>
+            <p style='color: red'><b> Please check your email and your spam folder ! </b></p
```
osallou, line 10: Could use bootstrap label warning for styling

FanchTheSystem: yes

FanchTheSystem:
image

PR #282, opened 2021-08-26
Allow admin to create account

FanchTheSystem:
The purpose of this pull request is to allow account creation when user does not have email adress.

The use case is at abims cluster, where imap account need ldap account to be created. So email does not exist when people register to My.

- add a link in admin menu to create a new account
- add a flag to allow setting of account when email does not exist (send mail copy to support email adress)


FanchTheSystem:
@osallou is it ok for you ?

FanchTheSystem:
So after dicussion, I ended with 3 checkbox, each one is explicit and set a flag with a revelant name.

image

PR #289, opened 2021-09-28
update package and package-lock

FanchTheSystem:
update redis, axios and amqplib

+ run npm audit fix

+ rollback lockfileVersion to 1

### DIFF
--- a/manager2/package-lock.json
+++ b/manager2/package-lock.json
@@ -1,17427 +1,8 @@
 {
   "name": "my-ui",
   "version": "0.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "my-ui",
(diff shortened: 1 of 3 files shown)


FanchTheSystem:
@osallou ?

osallou:
hum, did you test redis/amqplib ?  versions are really different, may break some stuff

FanchTheSystem:
Redis is tested on IFB and Abims dev env.
Rabbitmq is not tested as we don"t use it.

It look like both version update does not have any important breacking change, see:
- https://github.com/NodeRedis/node-redis/blob/master/CHANGELOG.md
v3 is not a real release, it look like it is only a report of all dev done (and breaking change does not concern us in my
- https://github.com/squaremo/amqp.node/blob/main/CHANGELOG.md
last breacking change was in v0.5.5 (we are already with v0.5.6)

PR #294, opened 2021-10-07
Feature multiexpirenotif

osallou:
on user expiration, send multiple notification emails

### DIFF
--- a/config/default.json.template
+++ b/config/default.json.template
@@ -50,6 +50,8 @@
             "series": "goacct.fixed.disk.panasas-omaha.*.user.#USER#"
         }
     },
+    "send_expiration_notif_to_admin": true,
+    "disable_auto_expiration": false,
     "plugin_script_dir": "/tmp/scripts",
     "user_extra_dirs": []
   },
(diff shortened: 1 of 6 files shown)


FanchTheSystem:
@osallou : maybe we should merge this will conflict are minor

osallou:
need to fix conflict

you really want to check all users expiration every day?

FanchTheSystem:
at ifb and abims we use  systemd timer to launch script, so modification on gomngr.sh will not have impact on our side

FanchTheSystem:
We plan to enable user and project expiration in few day/week.

So I think we will need an option to send mail to admin too.

And maybe a way to disable expiration the first month (only send mail, and don't change user password)

osallou:
if you want

FanchTheSystem:
conflict solved ;)

FanchTheSystem:
About config option, After reading the code. 

I think the second one is useless, the simplest way to do it is to disable the expiration is to disable the launch the expire.js script, (and it is easy to do it for ifb and abims as we use systemd timer)

FanchTheSystem:
https://github.com/genouest/genouestaccountmanager/pull/306

PR #302, opened 2021-11-12
Update my sync ldap -> mongo

FanchTheSystem:


FanchTheSystem:
@osallou my test are ok on this, I think you do not use this script at genouest, so maybe we can merge

PR #304, opened 2021-11-15
show user creation date in profile if availlable

FanchTheSystem:


core/user.service.js
```diff
@@ -51,6 +51,10 @@ async function activate_user(user, action_owner = 'auto') {
     if (!user.password) {
         user.password = Math.random().toString(36).slice(-10);
     }
+    if (!user.created_at) {
+        user.create_at = new Date().getTime();
```
osallou, line 55: created_at rather than create_at I suppose (typo)

osallou (changes requested):

manager2/src/app/user/user.component.html
```diff
@@ -4,8 +4,9 @@
           <h2><b>{{user.uid}}</b>
           <div *ngIf="subscribed && (session_user.is_admin || (config.enable_ui && config.enable_ui.newsletters))"><button class="btn" (click)="unsubscribe()"><small class="oi oi-envelope-open glyphicon glyphicon-envelope" title="The user is subscribed to the newletters" style="color:green"></small> unsubscribe</button></div>
           <div *ngIf="! subscribed"><button class="btn" (click)="subscribe()"><small class="oi oi-envelope-open glyphicon glyphicon-envelope" title="The user is not subscribed to the newletters" style="color:red"></small> subscribe</button></div>
-           -
-           <small>Expires: {{date_convert(user.expiration)}}</small></h2>
+          -
+          <small *ngIf="user.created_at">Created at : {{date_convert(user.created_at)}} - </small>
```
osallou, line 8: why a dash at the end of line?

core/user.service.js
FanchTheSystem: fixed 2d78dbf543c83518d3b72fefdd457acc0cbc506b

manager2/src/app/user/user.component.html
FanchTheSystem: because there are already a dash between newsletter and and "expires:"

And as I insert created between both, I added one dash.

But we can remove it if you want

PR #308, opened 2021-11-24
replace FIDO U2F auth by Webauthn as FIDO U2F is deprecated

osallou:


osallou:
support also google authenticator

PR #312, opened 2021-11-25
limit mail per min on expiration notif

FanchTheSystem:


cron/expire.js
```diff
@@ -3,6 +3,10 @@
 /**
  * Test expiration date of user, if expired, expire the user
  */
+
+const { promisify } = require('util');
```
osallou, line 7: where does this come from ?

util/promisify?

FanchTheSystem: https://nodejs.org/api/util.html#utilpromisifyoriginal

FanchTheSystem: in fact it allow to do a await on setimeout to simulate sleep
I have tested it and it work

PR #314, opened 2021-11-29
update package lock, to use lockfileversion 1

FanchTheSystem:
To avoid error like:
```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```

and run npm audit fix

### DIFF
--- a/manager2/package-lock.json
+++ b/manager2/package-lock.json
@@ -1,17327 +1,8 @@
 {
   "name": "my-ui",
   "version": "0.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "my-ui",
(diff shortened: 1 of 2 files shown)


osallou:
I do not see in package lock (at least) the otplib dep present in package.json

FanchTheSystem:
> I do not see in package lock (at least) the otplib dep present in package.json

line 111:
image

osallou:
Yes , removed....
But it should be present

FanchTheSystem:
yes you are right, maybe it is the diff which is not ok

because the line are present in the file submitted:
https://github.com/FanchTheSystem/genouestaccountmanager/blob/update_lockfile/package-lock.json#L2541-L2549

PR #315, opened 2021-12-02
Add some example to extend expiration on user and project

FanchTheSystem:


FanchTheSystem:
@osallou , I think you can merge this too, as there are no impact

PR #316, opened 2021-12-07
add an option to never change user home

FanchTheSystem:


osallou:
won't this affect scripts manipulating home ? (move home etc...)

FanchTheSystem:
> won't this affect scripts manipulating home ? (move home etc...)

yes ! the goal is to disable the move of home on abims cluster

without the variable in config **"never_update_user_home**, it should work as usual

osallou:
but when you change main group, you will not change user.home but script will still be executed (to chown etc...)

Template execs:

    oh=$(echo "{{ user.oldhome }}" | sed -e s:/$::g)
    nh=$(echo "{{ user.home }}" | sed -e s:/$::g)
    
    if [ "$oh" != "$nh" ]
    ...

oldhome will be the same than home in this case I suppose

FanchTheSystem:
> but when you change main group, you will not change user.home but script will still be executed (to chown etc...)
> 
> Template execs:
> 
> ```
> oh=$(echo "{{ user.oldhome }}" | sed -e s:/$::g)
> nh=$(echo "{{ user.home }}" | sed -e s:/$::g)
> 
> if [ "$oh" != "$nh" ]
> ...
> ```
> 
> oldhome will be the same than home in this case I suppose

Yes it will be the same so the if will not do anything

PR #317, opened 2021-12-08
add a cron to send project expiration mail to owner

FanchTheSystem:
copy of the test on user (with the 60/15/3 day counter) but for project owner


cron/test_projects.js
```diff
@@ -0,0 +1,100 @@
...
+            do_notify = true;
+        } else if (project.expiration_notif == 2 && (project.expire <= in3Days)) {
+            project.expiration_notif = 3;
+            do_notify = true;
+        }
+        if (!do_notify) {
+            continue;
+        }
+        console.log(`Project will expire, send notication number ${project.expiration_notif} to ${project.id}`);
+        let user = await dbsrv.mongo_users().findOne({uid: project.owner});
```
osallou, line 67: should manage case where user is not found

osallou (commented):
there is also a cron/test_quote_expire.js, badly named as it does not check quotas, but project expiration. Maybe it should just be replaced with this code (and call in cron/gomngr.sh)

cron/test_projects.js
FanchTheSystem: 701f4411fa1e46e0b91a3983e1a3048bda268613 Done, and I also add a check on project owner

FanchTheSystem:
> there is also a cron/test_quote_expire.js, badly named as it does not check quotas, but project expiration. Maybe it should just be replaced with this code (and call in cron/gomngr.sh)

Can't find test_quote_expire.js in the repo

osallou:
cron/project_quota_expire.js , typo error...

osallou:
should call this script in cron/gomngr.sh in this case in place of project_quota_expire

FanchTheSystem:
I have added the new script in the cron.

About project_quota_expire I will modify it in another PR to take action on project expiration (same kind of action we already have for user expiration)

osallou:
project_quota_expire had a wrong name, it simply test project expiration, so your script should just replace it (so delete it and replace in cron/gomngr.sh)

FanchTheSystem:
> project_quota_expire had a wrong name, it simply test project expiration, so your script should just replace it (so delete it and replace in cron/gomngr.sh)

done

cron/gomngr.sh
```diff
@@ -153,13 +145,21 @@ while true; do
...
-      rm /tmp/gomngr.lock
-      rm /tmp/gomngr.list
-      echo "Exit requested"
-      exit 0
+        rm /tmp/gomngr.lock
+        rm /tmp/gomngr.list
+        echo "Exit requested"
+        exit 0
+    fi
+    echo "Check for account upcoming expiration"
```
osallou, line 153: Project not account (and will be ok for merge)

PR #318, opened 2021-12-13
add option to crypt password in ldap if ldap not crypting itself

osallou:
* in goldap check if config.ldap.crypt is set, if yes, crypt as sha512 the passwords
* use in user.service a new_password function for password/key generations


osallou:
@FanchTheSystem oups, merged by accident. I made a few tests however, but if you still could check it looks fine for you

PR #329, opened 2022-01-20
allow admin tochange user groups when user is expired, Closes #324

### LINKED ISSUE #324: Changing user main group does not trigger any cron scripts when the user is expired.
No error is shown (and the user group is changed in the UI), but no cron scripts are written, thus the group in the LDAP stays the same.
This leads to an issue where the state of the user is different between the user manager and the LDAP.

To fix it after user activation, we need to switch back to the older group, and then back to the new one.

osallou:


routes/users.js
```diff
@@ -1645,15 +1653,18 @@ router.put('/user/:id', async function(req, res) {
...
         user.fid = fid;
-        if(user.oldemail!=user.email && !user.is_fake) {
-            await notif.modify(user.oldemail, user.email);
-        } else if(userWasFake && !user.is_fake) {
-            await notif.add(user.email);
-        }else if (!userWasFake && user.is_fake) {
-            await notif.remove(user.email);
+        // Change mail registration only when user is active
+        // as expired users are removed from mailing list
+        if(user.status == STATUS_ACTIVE) {
```
FanchTheSystem, line 1658: Maybe also allow email change for admin when user is expired

osallou: it is done on line 1496, here we just prevent mailing list registration (if user not active, should not add him to lists)

PR #346, opened 2022-03-17
prevent multiple click on the extend button

FanchTheSystem:


FanchTheSystem:
my test are ok for this one

PR #349, opened 2022-03-23
add never_expire flag for user

FanchTheSystem:


FanchTheSystem:
@osallou

manager2/src/app/user/user.component.html
```diff
@@ -173,11 +173,17 @@ <h4>Information</h4>
...
                           </div>
                           <div class="form-check row">
-                                <input [ngModelOptions]="{standalone: true}" [(ngModel)]="user.is_fake" type="checkbox" class="form-check-input">
+                              <input [ngModelOptions]="{standalone: true}" [(ngModel)]="user.is_fake" type="checkbox" class="form-check-input">
                               <label class="col-6 form-check-label">
                                   Fake/App account
                               </label>
                           </div>
+                          <div class="form-check row">
+                              <input [ngModelOptions]="{standalone: true}" [(ngModel)]="user.never_expire" type="checkbox" class="form-check-input">
```
osallou, line 182: If undefined (for all existing users), won t it fail?

On user.get should we not set never_expire to false?

FanchTheSystem: it work on my test, but I add it to get method to avoid any potential issue e47b318ecf6de6bc1c782906061df0b124589c6c

PR #352, opened 2022-03-28
protect ldap password with quote as it may contain some shell intepreted char

FanchTheSystem:


osallou:
why not but this is the ldap admin password, so managed by administrator themselves....

FanchTheSystem:
Yes of course, I have generate a strong password for the ldap admin on the ifb core installation (not the cluster one) 

and account creation failed because the ldap password contain shell special char

so the quote solve the issue

maybe one day you will change your ldap password to have a more secure one and will be happy to be able to have $ or @ inside

osallou:
but if you have multiple quote password ?  admin''stuff''' etc... ?

FanchTheSystem:
if you have one or more simple quote in your password it will not work as it is today as they will be interpreted by the shell 

maybe we should add a mention in the doc that ldap password with simple are not supported

osallou:
As better solution, password should be escaped

PR #362, opened 2022-06-02
Display expired projects

lecorguille:
This PR will display the "expired" project in the admin project page.

### DIFF
--- a/manager2/src/app/admin/projects/projects.component.html
+++ b/manager2/src/app/admin/projects/projects.component.html
@@ -202,3 +202,70 @@
     </p-table>
   </div>
 </div>
+
+<div class="card bg-light">
+  <div class="card-header bg-danger">
+      <h3>Expired [{{expired_projects.length}}]</h3>
+  </div>
+  <div class="card-body">
+    <div class="table-responsive table-striped">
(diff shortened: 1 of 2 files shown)


lecorguille:
I haven't tested so far.

osallou:
You do not change projects list.
Expired projects will appear both in projects and expired projects


When you get project list,  you could simply add to projects or expired projects depending on expiration notifs, this would be done in a single function and avoid 2 queries to list projects.

lecorguille:
Capture d’écran 2022-06-03 à 14 44 42

osallou:
cf mon premier commentaire , projets expirés appraitront tjs ds la liste des projets , et faut factoriser avec project_list

lecorguille:
> You do not change projects list.
> Expired projects will appear both in projects and expired projects

Since we don't have a real status... We can do that:
 - All expiration_notif < 3
 - Expired expiration_notif > 0

What do you think about that?

osallou:
you can use expiration notif, or just expire < Date.now()

lecorguille:
Indeed

lecorguille:
Ready to review and merge

PR #396, opened 2023-01-17
Add the possibility to give feedback to users based on their choice

KonoganB:
The goal is to give an immediate feedback to the user based on the type of data he wants to work with using the Genouest services when requesting an account. 
Using the config file, the admin can now give feedback using bootstrap alerts syntax.

### DIFF
--- a/config/test.json
+++ b/config/test.json
@@ -46,6 +46,12 @@
    "multiple": false
    },
    {
+    "title": "second choice with feedback",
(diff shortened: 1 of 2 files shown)


osallou:
@KonoganB ready for merge?

KonoganB:
> @KonoganB ready for merge?

Should be: doesn't seem to break anything, the test file was updated to show both possible behaviors of "radio" buttons (wasn't added to the multiple choice buttons as it isn't needed as of yet)

PR #408, opened 2023-10-18
Remove reservation owner link

lecorguille:
Remove reservation owner link


mboudet:
Should been fixed by #409

Issue #446, opened 2024-02-21
Add 'duration' selector on project request


mboudet:
Or an 'Expected end date' datepicker, for more flexibility. This should be viewable on the admin side when validating and listing 'waiting' projects

PR #459, opened 2024-04-25
Fix issue #446: Add 'duration' selector on project request 

rsiminel:
Is this sufficient for fixing the issue ?
As I understand, an admin is supposed to create each project manually anyway, copying the information off of the email they were sent. This means that now, they'll have to copy the expiration date manually as well. Is there a better way of doing things when add_project (from manager2/src/app/admin/projects.component.ts) is executed ?

### DIFF
--- a/manager2/src/app/admin/projects/projects.component.ts
+++ b/manager2/src/app/admin/projects/projects.component.ts
@@ -290,7 +290,7 @@ export class ProjectsComponent implements OnInit {
     }
 
     modify_project(project) {
-        this.new_project = project;
+        this.new_project = {...project, expire: new Date(project.expire)};
         this.update_project_on_event(project.id);
     }
 
--- a/manager2/src/app/admin/projects/projects.component.ts
+++ b/manager2/src/app/admin/projects/projects.component.ts
@@ -168,18 +168,19 @@ export class ProjectsComponent implements OnInit {
             return;
         }
         this.reset_msgs()
+        const project_to_send = { ...this.new_project, expire: new Date(this.new_project.expire).getTime()}
         this.projectService.add({
-            'uuid': this.new_project.uuid,
-            'id': this.new_project.id,
-            'expire': new Date(this.new_project.expire).getTime(),
-            'owner': this.new_project.owner,
-            'group': this.config.project.enable_group ? this.new_project.group : '',
-            'size': this.new_project.size,
-            'cpu': this.new_project.cpu,
-            'description': this.new_project.description,
-            'access': this.new_project.access,
-            'orga': this.new_project.orga,
-            'path': this.new_project.path
+            'uuid': project_to_send.uuid,
+            'id': project_to_send.id,
+            'expire': project_to_send.expire,
+            'owner': project_to_send.owner,
+            'group': this.config.project.enable_group ? project_to_send.group : '',
+            'size': project_to_send.size,
+            'cpu': project_to_send.cpu,
+            'description': project_to_send.description,
+            'access': project_to_send.access,
+            'orga': project_to_send.orga,
+            'path': project_to_send.path
         }).subscribe(
             resp => {
                 this.add_project_msg = resp.message;
--- a/manager2/src/app/admin/projects/projects.component.ts
+++ b/manager2/src/app/admin/projects/projects.component.ts
@@ -318,7 +318,7 @@ export class ProjectsComponent implements OnInit {
     modify_project(project) {
         const project_to_send = { ...project, expire: new Date(project.expire)}
         this.new_project = project_to_send;
-        this.update_project_on_event(project.id);
+        this.update_project_on_event(project_to_send.id);
     }
 
     reject_project(project) {
--- a/manager2/src/app/admin/projects/projects.component.ts
+++ b/manager2/src/app/admin/projects/projects.component.ts
@@ -150,7 +150,7 @@ export class ProjectsComponent implements OnInit {
         if (!this.new_project.expire) {
             this.new_project.expire = this.date_convert(new Date().getTime() + this.config.project.default_expire * this.day_time)
         }
-        
+
         if (!this.new_project.size || this.new_project.size == 0) {
             this.new_project.size = this.default_size;
         }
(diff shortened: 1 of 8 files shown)


mboudet (commented):
Seems fine on the 'request' side. However, we want the selected end date to be available on the admin side for validation.

On the [admin side](https://github.com/genouest/genouestaccountmanager/blob/364e137b3138e132539f3965618286ed0d179332/manager2/src/app/admin/project/project.component.html#L47), the column/parameter is called 'expire'. it need to match the column/parameter you are storing in DB (currently 'end')

manager2/src/app/project/project.component.html
```diff
@@ -29,6 +29,12 @@ <h3>Request a new project creation</h3>
               <small id="nameHelp" class="form-text text-muted"><i>Optional, for information only</i></small>
           </div>
 
+          <div class="form-group row col-sm-8">
+            <label for="new_project_end">Expected end date</label>
+            <input placeholder="project end date" type="date" id="new_project_end" [ngModelOptions]="{standalone: true}" [(ngModel)]="new_project.end" class="form-control"/>
+            <small id="nameHelp" class="form-text text-muted"><i>Optional, for information only</i></small>
```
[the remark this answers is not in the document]
rsiminel: Ah, I had not understood that. On it.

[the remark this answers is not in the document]
rsiminel: Did that do the trick ?
I'm not sure how the security over DataTime formating works in TypeScript.
I asume we should also modify line 26 in project.service.js in /core ?
How do we allow the admin validating the request to manually modify the expiration date in case the user entered something stupid ?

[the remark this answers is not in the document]
mboudet: This should work, yeah.
Before 'validating' a project, an administrator can modify all fields, so we can correct any issues with provided values.

manager2/src/app/admin/projects/projects.component.ts
```diff
@@ -175,15 +175,15 @@ export class ProjectsComponent implements OnInit {
         this.projectService.add({
             'uuid': this.new_project.uuid,
             'id': this.new_project.id,
+            'expire': this.new_project.expire,
```
mboudet, line 178: I 'think' you are going to need to convert date several times (in this file, but also on the user project file).

The issue is that angular datepicker give you a Date (a js entity type). But we don't want to store it in DB as it is, we want to store the unix time

(Using the `new Date(this.new_project.expire).getTime()` there was previously).
But then, you will need to convert it back to display it on the admin side (using something like [this](https://github.com/genouest/genouestaccountmanager/blob/364e137b3138e132539f3965618286ed0d179332/manager2/src/app/admin/project/project.component.ts#L190))

To summarize, the process would be:

* User select date in the field, and click send
* Before sending to the service, you convert the value using `new Date(this.new_project.expire).getTime()` (probably should not edit the variable in-place. Might be better to pass a whole new project Object with the values you want)
* Store the project in DB as you already do.
* When an admin access the projects admin page, get the project from DB, and convert it back to date from display

rsiminel: It should be good now.

mboudet: Still need to convert it to unix time on the 'request project' side, and convert it back to JS Date when displaying admin-side

rsiminel: Right, sorry for taking so long, I had to figure out what the heck a spread operator is.
Is there anything else I missed?

mboudet: Hmm, I think you still need to convert unix time to JS date on the admin side when getting the pending projets
Now that I'm thinking about it, you'll also need to manage the case when pending projects do not have an expire data (since current pending projects in production will not have one)

Also need to rollback package-lock.js and dev_quickstart/default.json I think?

rsiminel: Ok so Angular's [(ngModel)] directive handles conversion between Unix timestamps and date objects automatically so there is no need to do admin-side conversions to JS date types but we do have to convert back to unix timestamps manually.
Also, the GitLens extention for VS Code is fantastic. Here you are working on the same problem 5 years ago : https://github.com/genouest/genouestaccountmanager/pull/33

rsiminel: Ok, @mboudet, everything works now. The stupid html date input controls only accept a formated string (which is the original reason why the date_convert() function was written), which was causing me issues.

If there is a pending project with no expiration date, it will not cause any bugs but the admin accepting the project will have to enter one manually.

manager2/src/app/admin/projects/projects.component.ts
```diff
@@ -147,10 +147,6 @@ export class ProjectsComponent implements OnInit {
         // about order, see: https://medium.com/@lukaonik/how-to-fix-the-previous-ngmodelchange-previous-value-in-angular-6c2838c3407d
         this.new_project.id = tmpprojectid; // todo: maybe add an option to enable or disable this one
 
-        if (!this.new_project.expire) {
-            this.new_project.expire = this.date_convert(new Date().getTime() + this.config.project.default_expire * this.day_time)
-        }
-
```
mboudet, line 153: Why not keep this as a fallback if expire is not set (old pending projects), instead of being an empty string?

rsiminel: You're right, I brought it back.
While I'm at it, may I ask if [this](https://github.com/rsiminel/genouestaccountmanager/blob/56aca6f198bf2c5bce5a26d2099373424f103ff5/manager2/src/app/admin/project/project.component.ts#L92) is a sanctioned use of date_convert ?

mboudet: Not sure what you mean by 'sanctioned use' ?

rsiminel: Sorry, I was unclear. I meant to ask why the function is being called there and whether it might cause the date to be stored as a string in the database after that function call.

mboudet: I think 'show_project_users' is in charge of actually getting the project. Since we show the expiration date in a editable field, I believe it need to be converted from DB format in that step.

We convert it back before updating [here](https://github.com/genouest/genouestaccountmanager/blob/364e137b3138e132539f3965618286ed0d179332/manager2/src/app/admin/project/project.component.ts#L164)

rsiminel: Oh ok, thanks for explanation. I guess we can merge now?

mboudet:
Looks good to me. Does everything behave as expected?

mboudet (approved):

PR #465, opened 2024-05-16
Extended tp registrations

### LINKED ISSUE #453: Allow admin to extend Tp registrations

rsiminel:
I still have a couple questions that I included as comments in the code. While I'm at it, here are a couple more concerns :
While looking arround, I noticed that when a user wants to extend a Project, it is automatically extended by one year. Was I just supposed to do something like that ? Or should I make the Project extention more like this one ?
Also, should I add checks that TPs/Projects are not created retroactively (start/end date < current date) ?

closes #453

### DIFF
--- a/core/tps.service.js
+++ b/core/tps.service.js
@@ -236,7 +236,7 @@ async function delete_tp_users(users) {
 async function extend_tp_user(user, extension) {
     logger.debug('extend_tp_user', user.uid);
     try{
-        user.history.push({ 'action': 'reactivate', date: new Date().getTime() });
+        user.history.push({ 'action': 'extend', date: new Date().getTime() });
         await dbsrv.mongo_users().updateOne({ uid: user.uid }, {
             '$set': { expiration: extension.to }
         });
--- a/core/tps.service.js
+++ b/core/tps.service.js
@@ -253,6 +253,27 @@ async function remove_tp_reservation(reservation_id) {
 
 }
 
+async function extend_tp_reservation(reservation_id, new_expire) {
+
+    logger.info('Extend reservation', reservation_id);
+
+    let reservation = await dbsrv.mongo_reservations().findOne({'_id': reservation_id});
+    logger.debug('Extend reservation', reservation);
+
+    try {
+        await dbsrv.mongo_reservations().updateOne({'_id': reservation_id}, {
+            '$set': {
+                'to': new_expire
+            }
+        });
+        await dbsrv.mongo_events().insertOne({ 'owner': 'auto', 'date': new Date().getTime(), 'action': 'extend reservation for ' + reservation.owner , 'logs': [] });
+
+    } catch (error) {
+        logger.error(error);
+    }
+
+}
+
 async function create_tp_reservation(reservation_id) {
 
     logger.info('Create reservation', reservation_id);
--- a/core/tps.service.js
+++ b/core/tps.service.js
@@ -254,7 +254,7 @@ async function remove_tp_reservation(reservation_id) {
 
 }
 
-async function extend_tp_reservation(reservation_id, new_reservation) {
+async function extend_tp_reservation(reservation_id, extension) {
     logger.info('Extend reservation', reservation_id);
 
     let reservation = await dbsrv.mongo_reservations().findOne({'_id': reservation_id});
@@ -263,7 +263,7 @@ async function extend_tp_reservation(reservation_id, new_reservation) {
     try {
         await dbsrv.mongo_reservations().updateOne(
             {'_id': reservation_id},
-            {'$set': { 'to': new_reservation.to } }
+            {'$set': { 'to': extension.to } }
         );
         await dbsrv.mongo_events().insertOne({ 'owner': 'auto', 'date': new Date().getTime(), 'action': 'extend reservation for ' + reservation.owner , 'logs': [] });
 
--- a/core/tps.service.js
+++ b/core/tps.service.js
@@ -260,6 +260,10 @@ async function extend_tp_reservation(reservation_id, new_reservation) {
     let reservation = await dbsrv.mongo_reservations().findOne({'_id': reservation_id});
     logger.debug('Extend reservation', reservation);
 
+    if (new_reservation.to < reservation.to) {
+        // Error: not meant to advance expiration date
+    }
+
     try {
         await dbsrv.mongo_reservations().updateOne(
             {'_id': reservation_id},
--- a/core/tps.service.js
+++ b/core/tps.service.js
@@ -257,7 +257,6 @@ async function extend_tp_users(users, extension) {
 
 
 async function remove_tp_reservation(reservation) {
-    logger.info('Close reservation', reservation._id);
     logger.debug('Close reservation', reservation);
 
     if (reservation.accounts) {
@@ -289,7 +288,6 @@ async function remove_tp_reservation(reservation) {
 }
 
 async function extend_tp_reservation(reservation, extension) {
-    logger.info('Extend reservation', reservation._id);
     logger.debug('Extend reservation', reservation);
 
     try {
--- a/core/tps.service.js
+++ b/core/tps.service.js
@@ -97,7 +97,7 @@ async function extendExtraProject(project, extension) {
         return false;
     }
     try {
-        const extended_project = { ...project_to_extend, expire: extension.to }
+        const extended_project = { ...project_to_extend, expire: extension.to };
         let res = await prjsrv.update_project(project_to_extend.id, extended_project);
         return res;
     } catch(error) {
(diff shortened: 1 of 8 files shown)


mboudet:
> I noticed that when a user wants to extend a Project, it is automatically extended by one year. Was I just supposed to do something like that

Wow, I did not know this option existed (it's disabled on our instance). No it's fine to leave it that way, extension by an user and by an admin are two separate thing.

> Should I add checks that TPs/Projects are not created retroactively (start/end date < current date) ?

If it's not already implemented, sure.

--------------------

Now that I'm looking at the code, I think you'll need to extend the 'expiration' value of created TP account (if any) to match the new expiration date + the grace period

manager2/src/app/tps/tps.component.ts
```diff
@@ -174,6 +176,21 @@ export class TpsComponent implements OnInit {
         this.listEvents();
     }
 
+    extend_reservation(new_expire: Date) {
+        this.msg = '';
+        this.errmsg = '';
+        if (new Date(this.new_expire).getTime() < this.selectedEvent.end) {  // Why did we rename the 'to' field to 'end' for selected events (l.59) ?
```
mboudet, line 182: Not sure

manager2/src/app/tps/tps.component.html
```diff
@@ -93,13 +93,42 @@ <h3>Reservation</h3>
                     </table>
                 </div>
                 <button *ngIf="!selectedEvent.created && !selectedEvent.over && authorized" class="p-button p-button-sm p-button-secondary" (click)="cancel_reservation()">Cancel</button>
+                <!-- <button *ngIf="selectedEvent.created && authorized && !session_user.is_admin" class="p-button p-button-sm p-button-primary" (click)="ask_extend_reservation()">Ask for Extension</button> -->
```
mboudet, line 96: Maybe remove this?

routes/tp.js
```diff
@@ -339,4 +339,71 @@ router.put('/tp/:id/reserve/now', async function(req, res) {
...
+        isadmin = await rolsrv.is_admin(user);
+    } catch(e) {
+        logger.error(e);
+        res.status(404).send({message: 'User session not found'});
+        res.end();
+        return;
+    }
+
+    if(!user) {
+        res.status(404).send({message: 'User does not exist'});  // Why do none of the other routes give a 404 status here ?
```
mboudet, line 364: They probably should, but to be honest this is a real edge case: this mean someone is logged in, but somehow his account does not exists.

Not sure if this can actually happen

routes/tp.js
```diff
@@ -339,4 +339,71 @@ router.put('/tp/:id/reserve/now', async function(req, res) {
...
+    let reservation = await dbsrv.mongo_reservations().findOne(filter);
+    if(!reservation){
+        res.status(403).send({message: 'Not allowed to extend this reservation'});
+        res.end();
+        return;
+    }
+
+    if (req.body.to < reservation.to) {
+        res.status(403).send({message: 'Extended end date must be after current end date'});
+        // Should I add "res.end();" here ?
```
mboudet, line 392: Sure

routes/tp.js
```diff
@@ -339,4 +339,71 @@ router.put('/tp/:id/reserve/now', async function(req, res) {
...
+    try {
+        tpssrv.extend_tp_reservation(reservation_id, req.body);
+    } catch (error) {
+        logger.error(error);
+        res.status(500).send({message: 'Error while extending tp reservation'});
+        res.end();
+        return;
+    }
+
+    res.status(200).send({message: 'Reservation extended'});  // Why do none of the other routes give a 200 status on success ?
```
mboudet, line 405: it might be the default code sent

routes/tp.js
rsiminel: Is there a general rule for when we add `res.end();` or not ?

mboudet: Hmm, seems like res.send() calls res.end(), so no need for both. (But there are several place where we call both in the code, maybe it's an error or an old version behaviour?)

manager2/src/app/tps/tps.component.ts
mboudet: We could probably have a readme file somewhere with the various models I guess?

rsiminel:
> Should I add checks that TPs/Projects are not created retroactively (start/end date < current date) ? 

Done (using end date as cutoff).

> Now that I'm looking at the code, I think you'll need to extend the 'expiration' value of created TP account (if any) to match the new expiration date + the grace period

Done, I think. + extending any project or user accounts associated with the TP. @mboudet, could you check that I did it right ?

PS: Some of the files I'm working on are getting pretty long (500 lines +), would it be useful to try using [JavaScript Regions](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.JavaScriptRegions) ?

mboudet:
>PS: Some of the files I'm working on are getting pretty long (500 lines +), would it be useful to try using [JavaScript Regions](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.JavaScriptRegions) ?

This is a vscode specific comment I believe? No really a fan of editor-specific comments in the code.

Ps: [96e96d8](https://github.com/genouest/genouestaccountmanager/pull/465/commits/96e96d8dfde171c71c92c47f31e0fa99a620d769) broke the tests I think?

rsiminel:
Sorry about the broken tests, pesky semicolons ;)

PS: When a user does not fill in the "About" section when reserving a TP, the error message 'Tell us why you need some tp accounts' does not appear on the webpage until something is actually written in the "About" section. Shall I make a new issue for this or is it an easy fix?
Also, that error and the 'Quantity must be >= 1' error are only tested for in the backend.

manager2/src/app/tps/tps.component.ts
```diff
@@ -174,6 +180,25 @@ export class TpsComponent implements OnInit {
         this.listEvents();
     }
 
+    extend_reservation(new_expire: Date) {
```
mboudet, line 183: You're passing 'new_expire' here (and on the html side), but you are using this.new_expire instead.

I'm not sure if there is a point in passing it?

routes/tp.js
```diff
@@ -328,15 +310,79 @@ router.put('/tp/:id/reserve/now', async function(req, res) {
...
+        return;
+    }
+
+    let reservation_id = ObjectID.createFromHexString(req.params.id);
+
+    let filter = {};
+    if(isadmin) {
+        filter = {_id: reservation_id};
+    }
+    else{
```
mboudet, line 358: No need, since this cannot occurs

routes/tp.js
```diff
@@ -328,15 +310,79 @@ router.put('/tp/:id/reserve/now', async function(req, res) {
...
+        filter = {_id: reservation_id, owner: user.uid};
+    }
+    let reservation = await dbsrv.mongo_reservations().findOne(filter);
+    if(!reservation){
+        res.status(403).send({message: 'Not allowed to extend this reservation'});
```
mboudet, line 363: Since only admins can extend, this can only mean the reservation does not exists (with this ID). Should clarify the error message

routes/tp.js
```diff
@@ -328,15 +310,79 @@ router.put('/tp/:id/reserve/now', async function(req, res) {
...
     res.send({reservation: newresa});
-    res.end();
+});
+
+router.put('/tp/:id/reserve/extend', async function(req, res) {
+    if(! req.locals.logInfo.is_logged) {
+        res.status(403).send({message: 'Not authorized'});
+        return;
+    }
+    if(! sansrv.sanitizeAll([req.params.id])) {
```
mboudet, line 327: Might as well add a check for the 'to' parameter here, since you use it later

core/tps.service.js
```diff
@@ -206,13 +224,35 @@ async function delete_tp_user(user) {
...
             await delete_tp_user(user);
         }
     }
+}
+
 
+async function extend_tp_user(user, extension) {
+    logger.debug('extend_tp_user', user.uid);
+    try{
+        user.history.push({ 'action': 'extend', date: new Date().getTime() });
```
mboudet, line 239: You're pushing to user.history, but not updating it in DB

core/tps.service.js
```diff
@@ -246,11 +285,38 @@ async function remove_tp_reservation(reservation_id) {
...
         });
         await dbsrv.mongo_events().insertOne({ 'owner': 'auto', 'date': new Date().getTime(), 'action': 'close reservation for ' + reservation.owner , 'logs': [] });
+    } catch (error) {
+        logger.error(error);
+    }
+}
 
+async function extend_tp_reservation(reservation_id, extension) {
+    logger.info('Extend reservation', reservation_id);
+    let reservation = await dbsrv.mongo_reservations().findOne({'_id': reservation_id});
```
mboudet, line 295: You already pulled the reservation from DB in the api route, so maybe pass it here and reuse it instead of re-pulling it from the ID

rsiminel: In which case shall I also make the same change for '/tp/:id/reserve/now' and '/tp/:id/reserve/stop'?
They both also pull a reservation from the DB in the api route, pass the id to a service function that then re-pulls the reservation from the id.

mboudet: Sure, might as well save a few DB queries

manager2/src/app/tps/tps.component.html
```diff
@@ -95,11 +95,39 @@ <h3>Reservation</h3>
                 <button *ngIf="!selectedEvent.created && !selectedEvent.over && authorized" class="p-button p-button-sm p-button-secondary" (click)="cancel_reservation()">Cancel</button>
                 <button *ngIf="!selectedEvent.created && session_user.is_admin" class="p-button p-button-sm p-button-primary" (click)="create_reservation()">Create</button>
                 <button *ngIf="selectedEvent.created && !selectedEvent.over && session_user.is_admin" class="p-button p-button-sm p-button-warning" (click)="remove_reservation()">Remove</button>
+                <button *ngIf="selectedEvent.created && !selectedEvent.over && session_user.is_admin" type="button" class="p-button p-button-sm p-button-primary" data-toggle="modal" data-target="#extend">Extend</button>
```
mboudet, line 98: Can you switch this button position with the 'Remove' one? I feel like 'deletion' stuff should be last

mboudet (approved):

PR #468, opened 2024-05-28
Project description requirements

### LINKED ISSUE #462: Make project description field required with a minimal number of characters
#IFB

Some users do not fill in the project description field.

However with the concern over personal health data we need a description of the project to spot potential HDS issues.
This is therefore a request to make this field required.

rsiminel:
Adding front end checks and visual clues for project description requirements (length > 30 chars)
BobLamarley already covered back end checks.

closes #462

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.4.30 (TBA)
+
+* Added 'End date' to the projet request page
+* Set 'description' as a required field in the projet request page
+
 ## 1.4.29
 
 * Various security PR merged
--- a/manager2/src/app/project/project.component.html
+++ b/manager2/src/app/project/project.component.html
@@ -12,10 +12,11 @@
               <span style="color:red">(required)</span>
               <span class="glyphicon glyphicon-question-sign" title="Unique name for the project" tooltip></span>
             </label>
-            <input required placeholder="project name" type="text" maxlength="42" id="new_project_id" name="id" [(ngModel)]="new_project.id" class="form-control" #id="ngModel"/>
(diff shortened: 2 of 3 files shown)


mboudet:
Hmm, I think one of the PR already managed that, by setting 'ngNativeValidate', with mean that `minlength="30"` should be restricting it already. (Unless it doesn't work?)

rsiminel:
I pulled from the main branch after that PR and in my test env I was getting the "Project description length should at least be 30 char" error that is sent from the back end...

mboudet:
Ok, I'm guessing that ngNativeValidate set the form as invalid, but we do not check the form state before sending.

Not exactly sure how ngNativeValidate works when used with angular validation to be honest. Might be better to double check on angular side (especially if it does not works).

You might want to check if `this.new_project.description` exists before checking the .length, it might throw an error otherwise.

mboudet:
@BobLamarley when you made the PR for requiring 30 characters, did your implementation works on your side? It seems it does not restrict on our side, the query to the backend is sent no matter what

BobLamarley:
I don't know why it isn't working anymore with `ngNativeValide`, it was on my branch
Anyway, here is my pull request for fixing this : https://github.com/genouest/genouestaccountmanager/pull/469
Instead of using `ngNativeValidate` who use native HTML 5 form validation, i use angular form validation with `ngForm`

rsiminel:
The ng form validation works (but only once you've touched a field).
However, did you test this ? The back-end gives a "Project name empty" error. I don't think you can just get rid of the two-way binding of the new_project attributes.

mboudet:
Ok, so there is now two PR with mostly the same content. Which one should I review?

rsiminel:
> Ok, so there is now two PR with mostly the same content. Which one should I review?

This one. It includes the changes made in [Bob's other PR](https://github.com/genouest/genouestaccountmanager/pull/469) _and_ it works.

mboudet (approved):

PR #470, opened 2024-06-11
Db interface rework

### LINKED ISSUE #370: Optional database creation form
Allow a setting where instead of directly creating a DB, an user can 'request' a DB creation through a "project request"-like UI, including several questions (Like usage, resources, duration, other users using this DB...)

(Would be great if these information could be accessible in the UI for administrators after DB creation)

rsiminel:
This PR is exactly the same as #384, I just didn't want Konogan to get a bunch of notifications on an ancient PR from his previous employers. It seems to have been basically entirely finished.

Description: New database management workflow. Users now ask for a database and give relevant information, while the admins receive an email when a database is requested and can visualize pending requests and validate them as they see fit.

closes #370


routes/database.js
```diff
@@ -422,6 +598,57 @@ router.delete('/database/:id', async function(req, res) {
...
+
+    let filter = {name: req.params.id};
+    if(!session_user.is_admin) {
+        filter['owner'] = session_user.uid;
+    }
+
+    let pending_database = await dbsrv.mongo_pending_databases().findOne({name: req.params.id});
+    if(! pending_database || (pending_database.type!==undefined && pending_database.type != 'mysql')) {
+        await dbsrv.mongo_pending_databases().deleteOne(filter);
+        await dbsrv.mongo_events().insertOne({'owner': session_user.uid, 'date': new Date().getTime(), 'action': 'database request' + req.params.id + ' deleted by ' +  session_user.uid, 'logs': []});
```
rsiminel, line 638: missing white-space after request

routes/database.js
```diff
@@ -422,6 +598,57 @@ router.delete('/database/:id', async function(req, res) {
...
+        res.send({message: ''});
+        res.end();
+        return;
+    }
+    else {
+        await dbsrv.mongo_pending_databases().deleteOne(filter);
+
+        await dbsrv.mongo_events().insertOne({'owner': session_user.uid,'date': new Date().getTime(), 'action': 'database request' + req.params.id+ ' deleted by ' +  session_user.uid, 'logs': []});
```
rsiminel, line 646: missing white-space after request

rsiminel (commented):
This PR looked like it was basically done, just needed a few style modifications.
However, I'm getting a bunch of weird errors when testing out it's functionalities in the test environment:
- Email is not sent to admin upon DB request
- Databases stay in the "pending databases" list after creation
- Info from the "I will be the only one using this database" check-box is not shown anywhere else
- A few other oddities that I've included screenshots of

rsiminel:
Ok @mboudet, she's ready for your review. I still can't figure out the origin of the issue I mentioned with the "change database owner" feature.
A few other questions:
- What do we plan on doing with the "I will be the only one using this database" check-box information (because currently it isn't displayed anywhere at all) ?
- Should we send more emails to concerned parties over the lifetime of a Database (on creation by an admin, on deletion) ?
- Should we add an option for admins to delete other people's Databases directly ?

mboudet:
> Ok @mboudet, she's ready for your review. I still can't figure out the origin of the issue I mentioned with the "change database owner" feature. A few other questions:
> 
>     * What do we plan on doing with the "I will be the only one using this database" check-box information (because currently it isn't displayed anywhere at all) ?

Maybe we can add it in the admin view of the database, with a column 'Sole owner' or something. (If you can fix the column name while you're at it, I think it's missing a closing tag somewhere)

>     * Should we send more emails to concerned parties over the lifetime of a Database (on creation by an admin, on deletion) ?

We need to send a mail upon creating by an admin in all cases, because we need to send the DB password. (That's how we did beforehand anyway).
No need for the DB deletion, because we would need to contact the user beforehand anyway.

>     * Should we add an option for admins to delete other people's Databases directly ?
Sure, why not

manager2/src/app/database/database.component.html
```diff
@@ -0,0 +1,114 @@
+<div class="modal" style="z-index: 1500;" id="ask_database" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+
+      <div class="modal-header">
+        <h3>Request a new database creation</h3>
```
mboudet, line 6: Maybe just 'Request a new database'

test/my.js
```diff
@@ -661,6 +733,8 @@ describe('My', () => {
                 });
         });
 
+       
+
```
mboudet, line 737: Maybe add a test to make sure a user cannot create a DB (only admin)

routes/database.js
```diff
@@ -8,480 +8,382 @@ let my_conf = cfgsrv.get_conf();
...
+    if(!session_user) {
+        res.status(401).send({ message: 'Not authorized' });
         return;
     }
-
-    session_user.is_admin = isadmin;
-
-    if(!session_user.is_admin) {
-        res.status(401).send({message: 'Not authorized'});
+    if(is_admin) {
```
mboudet, line 45: Sould be a 'not' I think?

routes/database.js
```diff
@@ -8,480 +8,382 @@ let my_conf = cfgsrv.get_conf();
...
+        res.status(404).send({ message: 'User session not found' });
         return;
     }
-
-    if (!session_user){
-        res.status(401).send({message: 'Not authorized'});
+    if (!session_user) {
+        res.status(401).send({ message: 'Not authorized' });
         return;
     }
```
mboudet, line 111: I think this route is unprotected: ie, I can list databases of another user.
Not really a security risk, but maybe add a check if session_id == owner || is_admin

routes/database.js
```diff
@@ -8,480 +8,382 @@ let my_conf = cfgsrv.get_conf();
...
+    }
+    if (!req.body.expire) {
+        res.status(403).send({ message: 'No expiration date' });
         return;
     }
+    let pending_db = {
+        owner: req.body.owner ? req.body.owner : session_user.uid,
+        name: req.params.id,
+        type: req.body.type ? req.body.type : 'mysql',
+        host: req.body.host && sansrv.sanitize(req.body.host) ? req.body.host : CONFIG.mysql.host,
```
mboudet, line 152: Should be restricted to the default values I think. We don't want users requesting something other than mysql & the config.host

routes/database.js
```diff
@@ -8,480 +8,382 @@ let my_conf = cfgsrv.get_conf();
...
+        usage: req.body.usage ? req.body.usage : '',
+        size: req.body.size ? req.body.size : '',
+        expire: req.body.expire,
+        single_user: req.body.single_user !== undefined ? req.body.single_user : true
+    };
+    let create_db = !(req.body.create === false || (req.body.type && req.body.type !== 'mysql'));
```
mboudet, line 158: Should probably be true all the time

routes/database.js
```diff
@@ -8,480 +8,382 @@ let my_conf = cfgsrv.get_conf();
...
-        db_type = req.body.type;
+    if (!session_user) {
+        res.status(401).send({ message: 'Not authorized' });
+        return;
     }
-
-    let db_host = CONFIG.mysql.host;
-    if(req.body.host!=undefined && req.body.host && sansrv.sanitize(req.body.host)){
-        db_host = req.body.host;
+    if (req.body.owner && req.body.owner !== session_user.uid && !is_admin) {
```
mboudet, line 213: Should just check is_admin here. We don't want users creating DB directly.

routes/database.js
```diff
@@ -8,480 +8,382 @@ let my_conf = cfgsrv.get_conf();
...
-                    '#DBUSER#': req.params.id,
-                    '#DBPASSWORD#': password
-                });
-            } catch(error) {
-                logger.error(error);
-            }
 
+        await dbsrv.mongo_databases().insertOne(db);
+        let filter = { name: req.params.id };
+        if (!is_admin) {
```
mboudet, line 251: Should not happen

routes/database.js
```diff
@@ -8,480 +8,382 @@ let my_conf = cfgsrv.get_conf();
...
+            filter['owner'] = session_user.uid;
+        }
+        await dbsrv.mongo_pending_databases().deleteOne(filter);
```
mboudet, line 254: I think you already delete the pending db in the create_db function (which is better, because it's only deleted after creating the db)

The case '!create_db' should only happen when a admin declare a DB, so nothing should need to be deleted.

routes/database.js
```diff
@@ -8,480 +8,382 @@ let my_conf = cfgsrv.get_conf();
...
+        'logs': []
+    });
+    res.send({ message: 'Database removed' });
+    return;
 });
 
 
-router.delete_dbs = async function(user){
-    let databases = await dbsrv.mongo_databases().find({'owner': user.uid}).toArray();
+router.delete_dbs = async function(user) {
```
mboudet, line 376: Maybe move this to core (and modify the import in tps.service.js)

mboudet:
As an additional comment: Currently `router.post('/database/:id'` serves for both DB creation, and DB declaration (with both restricted to admins, or at least should).

Might be better to have a dedicated route for each (like /database/create/:id/and /database/declare/:id), which would avoid having some convoluted test to decide if we need to create it or not.

manager2/src/app/admin/databases/databases.component.ts
```diff
@@ -58,33 +64,131 @@ export class DatabasesComponent implements OnInit {
...
+                );
             },
-            err => this.err_msg = err.error.message
-        )
+            err => { this.err_msg = err.error.message; }
+        );
     }
 
+
+    pending_list(refresh_requests = false) {
```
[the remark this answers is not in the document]
rsiminel: You're right, every time that function is called, it is called without an argument. I got rid of the refresh_requests part.

mboudet (approved):

PR #472, opened 2024-06-28
Revamp tp calendar

### LINKED ISSUE #467: Overhaul calendar

rsiminel:
closes #467


rsiminel:
@mboudet

Events are given a color corresponding to whether they are pending, active or over.

The events are shown as bars instead of dots now but there is no guarantee that the bars corresponding to one event are contiguous. Even worse, in some cases the bar of one event can seem to lead right into the bar of another separate event. Shall I write some logic to fix this ?

Once everything looks good, shall I write all of the styling as proper CSS ?

PR #473, opened 2024-07-01
Project back button

rsiminel:
Just adding a button to go back to the projects overview page when an admin is looking at an individual project.


rsiminel:
@mboudet

mboudet:
Can you re-submit this PR with only the changes you want to merge? There is way too much noise in this one (980 lines modified?)

Also, please add your changes to the changelog

PR #479, opened 2024-07-08
Revamp tp calendar

### LINKED ISSUE #467: Overhaul calendar

rsiminel:
@mboudet

The events are shown as bars instead of dots now but there is no guarantee that the bars corresponding to one event are contiguous.
Events are given a color corresponding to whether they are pending, active or over with a little bit of semi-random variation to distinguish events from one another. Perhaps a better "UX engineer" than I can come up with sexier colors and a better way to introduce their semi-randomness.
Shall I write up the styling in proper CSS or leave it as is ?

closes #467

### DIFF
--- a/manager2/src/app/tps/tps.component.ts
+++ b/manager2/src/app/tps/tps.component.ts
@@ -7,9 +7,9 @@ import { CalendarEvent, CalendarEventTimesChangedEvent } from 'angular-calendar'
 import { Subject } from 'rxjs';
 
 const eventColors = {
-    created: { primary: '#00dd00' },
-    pending: { primary: '#00aabb' },
-    over: { primary: '#0000ff' }
+    created: { primary: '#00dd44' },
+    pending: { primary: '#0066ff' },
+    over: { primary: '#808080' }
 };
 
 @Component({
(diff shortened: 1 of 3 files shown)


manager2/src/app/tps/tps.component.ts
```diff
@@ -48,17 +53,27 @@ export class TpsComponent implements OnInit {
...
+        let baseColor;
+        if (over) { baseColor = eventColors.over; }
+        else if (created) { baseColor = eventColors.created; }
+        else { baseColor = eventColors.pending; }
+        let color = { ...baseColor }
+        const hash = id.split('').reduce((acc, char) => acc + char.charCodeAt(0), 0);
+        const red = (hash % 128).toString(16).padStart(2, '0');
+        color.primary = `#${red}${color.primary.slice(3)}`;
+        return color;
+    }
```
mboudet, line 66: I fell like the color should be variable only for 'active' event. We can keep the same color for all the other events.
How variable are the colors?

mboudet: On askomics we use [this](https://github.com/askomics/flaskomics/blob/dev/askomics/react/src/classes/utils.jsx#L53) to generate the color, not sure if it's 'better'

rsiminel: Here is the way it looked before:
TpCalendar1

Here it is with the color scheme you are recommending:
TpCalendar2

This is way better indeed. Any other improvements ?

mboudet (approved):

PR #480, opened 2024-07-10
Always select newly created groups

### LINKED ISSUE #444: Creating a new group via user page does not always select the created group.
When creating a new group through the shortcut button on the user page, this group should be selected automatically, but it's not always the case.

Need to investigate to reproduce the issue.

rsiminel:
@mboudet
closes #444

The visible manifestation of the problem was that immediately after group creation using the shortcut button on the user page, the drop-down selector would display the first group in the list, despite everything being correct internally.

All that was needed was to use `[ngValue]` instead of `[value]` to bind the group names in the `<option>` elements of the selector.
Indeed, according to the Angular documentation, the `ngValue` directive "is mainly used on [...] option elements, so that when the element is selected, the [ngModel](https://docs.angularjs.org/api/ng/directive/ngModel) of that element (or its [select](https://docs.angularjs.org/api/ng/directive/select) parent element) is set to the bound value."

This has been by far the most time I've ever spent to fix a bug by changing just three characters.


mboudet:
@rsiminel Please add to the changelog and I'll merge

mboudet (approved):

mboudet:
Thanks for the cleanup!

PR #481, opened 2024-07-17
Add User, Project and Group classes

rsiminel:
@mboudet

I verified that most functionalities still work and everything seems to still be displayed correctly. Is there some rigorous way to make sure these heavy modifications have not broken anything?


lecorguille:
François Gerbes implemented some Selenium tests.
I don't know if there are still available somewhere?

mboudet:
I'm not sure if they were ever on this repo 🤔

manager2/src/app/user/user.component.ts
```diff
@@ -352,13 +335,13 @@ export class UserComponent implements OnInit {
             }
         }
         if (!found) {
-          this.groups.push({name: this.user.group, new: true});
-          this.missing_group = this.user.group;
+          this.groups.push(this.groupService.mapToGroup({ name: this.user.group, new: true }));
+          this.group_exists = false;
```
mboudet, line 339: I 'think' the reason I used `this.missing_group = this.user.group;` and then an `if user.group == this.missing_group` was that if you switch the selected group, you want the 'create group' button to go away.

By setting a static true / false variable, the button will always stay even if I select an existing group

mboudet (commented):
Just for clarification: I think you are using user.temp as a placeholder for various 'runtime' variables. Is that the case?

As far as I can see, we only use it for the 'authorized' key. Or is there any other variable we store in there?

manager2/src/app/user/user.component.ts
rsiminel: You're right. Modification reverted.

rsiminel:
> Just for clarification: I think you are using user.temp as a placeholder for various 'runtime' variables. Is that the case?
> As far as I can see, we only use it for the 'authorized' key. Or is there any other variable we store in there?

Yes exactly.
We store the 'authorized' boolean for use in `groups.component` and the 'access' boolean for use in `project.component`.

mboudet (approved):
Ok, a quick test on my end seems fine. Merging.

PR #482, opened 2024-07-17
better admin project selection

### LINKED ISSUE #438: Cannot create new project from admin UI without selecting a pending one
In Admin -> Projects, trying to directly create a new projet with the 'Project creation' form (without clicking on 'View' on a pending project first) will fail. This is because the workflow relies on a pending_project.uuid. (From the database).

Should 

* Disable the 'Save changes' button if the project does not have an uuid
* When clicking on 'Create project', actually use the 'add_project' method if there are no uuid

rsiminel:
@mboudet 
closes #438 

Most of the issue seems to have already been resolved. The `add_project` method is already used to create new projects even if there they have no uuid and in my test environment, new projects seems to to be created correctly even when none have been selected by the admin.

All I had to add were some esthetic modifications: the "Save changes" button only appears when a pending project has been selected and I added an "Undo selection" to de-select the selected project request if the admin wants to create a new project after having selected a pending project.

Let me know if I missed anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Add back button from single project view
 * Updated TP calendar to increase readability
 * Fix newly created group selection bug
+* Update admin project creation UI: disappearing buttons and add "undo selection" button
 
 ## 1.4.29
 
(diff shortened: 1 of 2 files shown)


mboudet:
Add the change to the changelog and I'll merge

mboudet (approved):

PR #541, opened 2024-10-29
Tp expiration

### LINKED ISSUE #499: In TP reservation, allow to set an 'expiration' date
There are several use case where we might want the TP accounts to be unaccessible, but not deleted.
(Ex: Homework cutoff date).

Would be good to be able to set an 'expire' date where all the TP accounts would be expired, but not deleted.

rsiminel:
@mboudet 
Will close #499 once finished.

In the [notification email after TP creation](https://github.com/rsiminel/genouestaccountmanager/blob/tp-expiration/core/tps.service.js#L193), how should we differentiate between the "expiration" from `CONFIG.tp.extra_expiration` and the "expiration" field that we are adding in this PR?
How should we handle a TP coming to the end of it's expiration date?
Should we just add a button for a TP owner or an admin to make the TP accounts inaccessible to students, like a "Lock" button that activates a "locked" state?

### DIFF
--- a/core/tps.service.js
+++ b/core/tps.service.js
@@ -288,11 +288,16 @@ async function remove_tp_reservation(reservation) {
 
 async function extend_tp_reservation(reservation, extension) {
     logger.debug('Extend reservation', reservation);
-
     try {
-        await dbsrv.mongo_reservations().updateOne({ '_id': reservation._id }, {
-            '$set': { 'to': extension.to }
-        });
+        if (extension.expire) {
+            await dbsrv.mongo_reservations().updateOne({ '_id': reservation._id }, {
+                '$set': { 'to': extension.to, 'expire': extension.expire }
+            });
+        } else {
+            await dbsrv.mongo_reservations().updateOne({ '_id': reservation._id }, {
+                '$set': { 'to': extension.to }
+            });
+        }
         await dbsrv.mongo_events().insertOne( {
             'owner': 'auto',
             'date': new Date().getTime(),
--- a/core/tps.service.js
+++ b/core/tps.service.js
@@ -389,12 +389,13 @@ async function create_tp_reservation(reservation_id) {
     }
 }
 
-async function tp_reservation(userId, from_date, to_date, quantity, about, group_or_project, name) {
+async function tp_reservation(userId, from_date, to_date, expire_date, quantity, about, group_or_project, name) {
     // Create a reservation
     let reservation = {
         'owner': userId,
         'from': from_date,
         'to': to_date,
+        'expire': expire_date,
         'quantity': quantity,
         'accounts': [],
         'about': about,
--- a/manager2/src/app/tps/tps.component.ts
+++ b/manager2/src/app/tps/tps.component.ts
@@ -40,6 +40,7 @@ export class TpsComponent implements OnInit {
     quantity: number
     fromDate: Date
     toDate: Date
+    expirationDate: Date
     about: string
     authorized: boolean
 
@@ -48,7 +49,7 @@ export class TpsComponent implements OnInit {
 
     activeDayIsOpen: boolean = true;
 
-    new_expire: Date
+    new_toDate: Date
 
     constructor(
         private authService: AuthService,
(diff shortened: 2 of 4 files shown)


mboudet:
Yeah, 'expiration' is the wrong term I guess, maybe 'lock' would be better (though effectively, the accounts would be expired in our system).

I feel like a 'lock' button available user-side would need an 'unlock' button to avoid mistakes :thinking: .

The cleanest way would be a 'lock date' on the TP registration field. On the lock date we expire all TP accounts (but do not delete them)

We also need to manage passing this info through the mail sent to the TP manager, which might be a bit complicated (since replacing variable is not very flexible). Maybe setting #LOCKED# to '' by default, and to 'Account will be locked on this date' if set? That would leave an empty line in the mail if the value is not set, but it's acceptable.

/!\ This mean that in the deletion workflow, we need to account for the case that the TP account might already be expired (making sure to not throw an error)

manager2/src/app/tps/tps.component.ts
```diff
@@ -149,22 +153,24 @@ export class TpsComponent implements OnInit {
...
                 return;
             }
             if(fromDate > toDate) {
                 this.reserrmsg = 'End date must be superior to start date';
                 return;
             }
-            if(toDate < new Date().getTime()) {
-                this.reserrmsg = 'End date can not be in the past';
+            if(toDate < new Date().getTime() || lockDate < new Date().getTime()) {
+                this.reserrmsg = 'End or lock date can not be in the past';
```
mboudet, line 166: Should also check that lock date <= end date

manager2/src/app/tps/tps.component.ts
```diff
@@ -225,15 +231,16 @@ export class TpsComponent implements OnInit {
...
         this.msg = '';
         this.errmsg = '';
-        if(new Date(this.new_expire).getTime() < this.selectedEvent.meta.end) {
+        const new_to_Date = new Date(this.new_toDate).getTime();
+        if(new_to_Date < this.selectedEvent.meta.end) {
             this.errmsg = 'Extended end date must be after current end date';
             return;
         }
-        if(new Date(this.new_expire).getTime() < new Date().getTime()) {
+        if(new_to_Date < new Date().getTime()) {
```
mboudet, line 239: Need a check for lock_date too (both cannot be in the past, and must be <= end date)

core/tps.service.js
```diff
@@ -192,11 +193,11 @@ async function send_user_passwords(owner, from_date, to_date, users, group) {
             '#FROMDATE#':  from.toDateString(),
             '#TODATE#':  to.toDateString(),
             '#EXPIRATION#': CONFIG.tp.extra_expiration,
+            '#LOCKED#': lock.toDateString(),
```
mboudet, line 196: Not sure how we want to manage this. As it stand, we will print the locked date everytime (even if its equal to the end date), which might be confusing to the user.

Maybe instead of passing a date directly, optionally passing a full string?
That way, if there is not lock date, there would just be a blank line in the mail

mboudet:
Three more things:

* How do we want to manage the 'extra_expiration' ? I'm guessing that if the user provided a lock date, we want to actually lock in at that date, and ignore the extra_expiration.

* Currently, the 'lock' key does not do anything. You need to modify cron/reservation_remove.js to take it into account.
(Need to make sure not to lock if lock_date == end_date (without the extra expiration)

* We need to manage the case where the users are locked, but the reservation owner extend the lock date beyond today.
(By unlocking the users I'm guessing?)

core/tps.service.js
rsiminel: Currently, we display the lock date in the TP overview every time, like this:
image

I don't think it's too bad to display the lock date everywhere. If you think this is important, should we then have "lock" be set to null when the user does not specify a lock date instead of setting it to be equal to the end date?

mboudet: I was mostly talking about the email sent with the passwords:  like [here](https://github.com/genouest/genouestaccountmanager/blob/master/config/default.json.template#L379C6-L379C18).

I'm not sure if adding a line with 'Accounts will be locked on #LOCK' make sense if a user did not select a lock time.

The issue is also that if the user did not select a lock date, it will seems like the accounts will be locked on the end date, and then be deleted 5 days later (the grace period). Might be confusing since they did not ask for a lock.

PR #546, opened 2024-11-13
start all service including back and front when doing docker compose

BobLamarley:
start all service including back and front when doing docker compose

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Clearer error message when adding a website
 * Fix error in 'projects' tab from user page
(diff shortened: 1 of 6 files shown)


BobLamarley:
I've changed that, i just have a problem with ldap config when having the frontend and backend inside a container instead of host machine, I think it's just a problem with localhost or ldap ( the container name )
I tried with changing the host inside the default.json, as I've done for the mongo:
image
But it wasn't the solution

Also there was already a dockerfile at the base of the project that is starting the node and frontend, but I think this one used for production so I've recreated one Dockerfile.dev for the node app at the root of the project and a Dockerfile for starting the front in manager2/

PR #564, opened 2024-12-09
Username accents

### LINKED ISSUE #501: Accentuation issues when updating an existing user
Example: I can register with the surname 'léa'.
However, as an admin, I cannot modify the user account because 'léa' contains unauthorized characters.

(This mean i cannot update the group/email/etc until I set the surname to 'lea')

rsiminel:
closes #501 
@mboudet 

Not sure if this actually resolves the issue as-is, but these are the only places in the code where we verify the contents of user's first or last names so it should be good.


mboudet (approved):

mboudet:
Looks fine to me

PR #565, opened 2024-12-09
add project activation email

### LINKED ISSUE #460: Send an email on project activation
To let the user know its project is ready to use

rsiminel:
closes #460 

I'm assuming we want to include more info than just this in the email we send?


mboudet:
Woops. Can you also add the mail content to the 'default.json.template', so that new deployments will not have any issues?

mboudet (approved):

PR #566, opened 2024-12-09
check project owner is not removed

### LINKED ISSUE #498: Removing a user who is a project owner should raise a warning / error

rsiminel:
closes #498 
@mboudet 

Yup, all that was needed was a simple check that the user being deleted isn't the project owner (no need to check if it's the last member of the project because in that case they would be the project owner anyway).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * add a readonly input field under email for showing the email domain*
 * Add "custom_users" key to config file, to be used with various scripts
 * Check that user being removed is not owner or last member of a project
+* Add email sent to a project owner when their project is activated
 * Separate global overview from individual group view in admin group management
 * Allow accents in user's first and last names
 
--- a/config/default.json.template
+++ b/config/default.json.template
@@ -315,6 +315,18 @@
      "<p>You will receive an email when your account has been validated.</p>",
      "<hr>"
     ],
+    "activated_project": [
+      "Your requested project '#NAME#' has been approved and activated by an admin in My.",
+      "It is for members of the '#GROUP#' group, you can find it at the following path:",
+      "'#PATH#'",
+      "Congratulations!"
+    ],
+    "activated_project_html": [
+      "<p>Your requested project '#NAME#' has been approved and activated by an admin in My.</p>",
+      "<p>It is for members of the '#GROUP#' group, you can find it at the following path:</p>",
+      "<p>'#PATH#'</p>",
+      "<p>Congratulations!</p>"
+    ],
     "add_to_project": [
         "#UID# has been added to project '#NAME#' in My",
         "#DESC#",
--- a/config/test.json
+++ b/config/test.json
@@ -255,6 +255,18 @@
      "<p>You will receive an email when your account has been validated.</p>",
      "<hr>"
     ],
+    "activated_project": [
+      "Your requested project '#NAME#' has been approved and activated by an admin in My.",
+      "It is for members of the '#GROUP#' group, you can find it at the following path:",
+      "'#PATH#'",
+      "Congratulations!"
+    ],
+    "activated_project_html": [
+      "<p>Your requested project '#NAME#' has been approved and activated by an admin in My.</p>",
+      "<p>It is for members of the '#GROUP#' group, you can find it at the following path:</p>",
+      "<p>'#PATH#'</p>",
+      "<p>Congratulations!</p>"
+    ],
     "add_to_project": [
         "#UID# has been added to project '#NAME#' in My",
         "#DESC#",
--- a/core/project.service.js
+++ b/core/project.service.js
@@ -54,6 +54,25 @@ async function create_project(new_project, uuid, action_owner = 'auto') {
     catch(error) {
         logger.error(error);
     }
+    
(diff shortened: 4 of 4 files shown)


mboudet:
> (no need to check if it's the last member of the project because in that case they would be the project owner anyway).

Not so sure about that. Currently, there is no check on the user deletion route regarding projects, and the project 'owner' is stored on the project itself. -> If we delete a user who is a project owner, it's still going to be the owner. So if there is another member of the project, it's going to be the last member of the project, but not the owner.

So, what is needed is a check on the 'user delete route': if the user we are trying to delete is the owner of the project, OR the last user, we abort and send an error message.

There is already a check for websites and DBs here: https://github.com/genouest/genouestaccountmanager/blob/master/routes/users.js#L529 , might as well add a check for project ownership.

rsiminel:
> If we delete a user who is a project owner, it's still going to be the owner. So if there is another member of the project, it's going to be the last member of the project, but not the owner.

So you're describing a case where the project owner's account was deleted before this PR?
Projet membership is stored on the user, so we'd have to look at every single user to see if one of them is in the same project.

Might I suggest 3 options: 
- we load every single User from the db every time we delete an account (I'm assuming this is sub-optimal)
- we add a member_count attribute to the Project class
- we check that all project owners correspond to existing accounts (I prefer this solution)

mboudet:
> Might I suggest 3 options:
> 
>     * We load every single User from the db every time we delete an account. I'm assuming this is sub-optimal.

Not THAT suboptimal, honestly. We don't delete users often (and we already do it [for the groups(https://github.com/genouest/genouestaccountmanager/blob/54af62db1297bcc8c783d6d75e6c05661788239e/core/group.service.js#L129) when deleting an user).
Plus we do this query anytime we go to the admin page for a project anyway. A proper mongo query filtered on project membership is fairly fast.

>     * We add a member_count attribute to the Project class.

Not really fond of this one

>     * We add a check that all project owners correspond to existing user accounts. I prefer this solution in a 'two birds, one stone' kind of way.

Where would this check be, and how would it run? (And how would it communicate the issue to the admin?)

mboudet (dismissed):

routes/users.js
```diff
@@ -536,6 +536,19 @@ router.delete('/user/:id', async function(req, res) {
...
             return;
         }
+        let projects = await dbsrv.mongo_projects().find({ owner: uid }).toArray();
+        if (projects && projects.length > 0) {
+            res.status(403).send({ message: 'User owns some projects, please change owner first!' });
+            return;
+        }
+        const allprojects = user.projects ? user.projects : [];
+        for (let project_name of allprojects) {
+            const users_in_project = await dbsrv.mongo_users().find({ 'projects': project_name }).toArray();
```
mboudet, line 546: Might be better to use .count() instead of toArray, and just check the count value after

mboudet: (Need to check if it works. It works in mongo directly, but well..)

routes/users.js
```diff
@@ -536,6 +536,19 @@ router.delete('/user/:id', async function(req, res) {
...
+            if (users_in_project.length <= 1) {
+                res.status(403).send({ message: 'User is the last member of project ${project_name}' });
```
mboudet, line 548: Might be better to store project names somewhere, and send the response after will all projects.
(Otherwise, you'll try to delete -> change a project -> try to delete -> etc)

routes/users.js
```diff
@@ -542,13 +542,17 @@ router.delete('/user/:id', async function(req, res) {
...
-            if (users_in_project.length <= 1) {
-                res.status(403).send({ message: 'User is the last member of project ${project_name}' });
-                return;
+            const users_in_project = await dbsrv.mongo_users().find({ 'projects': project_name }).count();
+            if (users_in_project <= 1) {
+                empty_projects += ', ' + project_name;
             }
         }
+        if(empty_projects != '') {
+            res.status(403).send({ message: 'User is the last member of project(s)${empty_projects}' });
```
mboudet, line 555: Missing a space in the error message.

routes/users.js
```diff
@@ -542,13 +542,17 @@ router.delete('/user/:id', async function(req, res) {
...
         const allprojects = user.projects ? user.projects : [];
+        let empty_projects = '';
         for (let project_name of allprojects) {
-            const users_in_project = await dbsrv.mongo_users().find({ 'projects': project_name }).toArray();
-            if (users_in_project.length <= 1) {
-                res.status(403).send({ message: 'User is the last member of project ${project_name}' });
-                return;
+            const users_in_project = await dbsrv.mongo_users().find({ 'projects': project_name }).count();
+            if (users_in_project <= 1) {
+                empty_projects += ', ' + project_name;
```
mboudet, line 551: Maybe you can add the projet names into an array, and use array.join(", ") to get an usage string.
(Would avoid the trailing ",")

routes/users.js
rsiminel: Currently displays : "User is the last member of project(s), project1, project2"

mboudet: ah right, you were setting the comma before.

mboudet (approved):

PR #567, opened 2024-12-09
Increase width UI#562

BobLamarley:
Ui take full screen without padding, not theme dependent, it's global styles.css
image


BobLamarley:
@mboudet this one can be reviewed

mboudet:
So, this contains a `Temporary fix before trying to remove assets/css/*.css files`

Is it ready as it is?

PR #572, opened 2024-12-18
Group view

rsiminel:
Makes the admin groups management more like the way it is for projects. There is a general Groups page where admins can create new groups and see a list of all groups. This list links to individual Group pages, where-from an admin can manage and modify information about one particular group.

Also made deleting a group more secure with a my-delete-confirm button.


mboudet:
Can you fix #580 at the same time?

mboudet:
Minor graphical stuff, but maybe we could set the 'owner' and 'description' fields inline in the new view?
(Also, centering the 'update' button maybe)

mboudet (approved):

Issue #555, opened 2024-11-29
Automatically delete users  account requests in stage "pending email approval". after X days 


roadn:

BobLamarley:
Do you want a bash script with a crontab that you launch background and do this job ?

roadn:
> Do you want a bash script with a crontab that you launch background and do this job ?

yes . Thank you. With email notification

BobLamarley:
Email notification to who ? The admin with a summary of all users deleted ? To each users himself ? To admin + users ?

roadn:
To admin + users

BobLamarley:
What i've done:
- i search for all the users already pending after than x days you provide in the command arg
- a mail to the user concerned + ONE mail to the admin that recap all users deleted

the script is a node script instead of bash to use `maisrv` functions to send mail
you have to specify first :
`export NODE_CONFIG_DIR=path/to/config`
Then usage is `node delete_pending_email_approvals.js --days 30` if you run it manually
You can also crontab this script with something like `* * * * * /path/to/node /path/to/delete_pending_email_approvals.js --days 30`

PR #576, opened 2024-12-30
Automatically delete users account requests in stage "pending email approval". after X days #555

BobLamarley:
### Origin: 
#555 

### What I’ve done:
- The script searches for all users who have been in the “pending approval” status for more than the specified number of days (provided via the command argument).
- It sends an email to each affected user and a single summary email to the admin, listing all the users who were deleted.

### Key Details:
- This is a **Node.js** script rather than a Bash script, as it leverages `maisrv` functions to send emails.
- To run the script, you’ll need to set the `NODE_CONFIG_DIR` environment variable to point to the configuration directory:

    ```bash
    export NODE_CONFIG_DIR=path/to/config
    ```

- **Usage**: To run the script manually, use the following command:

    ```bash
    node delete_pending_email_approvals.js --days 30
    ```

- **Scheduling**: You can also schedule this script using `cron`. For example, to run it every minute:

    ```bash
    * * * * * /path/to/node /path/to/delete_pending_email_approvals.js --days 30
    ```


mboudet:
I'd rather avoid adding deps if we can avoid it. Can you use commander as a CLI tools? We already use it in the various scripts in the bin folder.

tools/delete_pending_email_approvals.js
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        const deletedUsers = [];
+
+        for (const user of users) {
+            //await users_db.remove({ _id: user._id });
+            console.log(`User with UID ${user.uid} deleted.`);
+
+            const userMessage = `Your account has been deleted due to inactivity for over ${DAYS_TO_DELETE} days. If this was an error, please contact support.`;
+            await notify([user.email], 'Account Deletion Notice', userMessage);
```
mboudet, line 70: I wouldn't send a mail to user who could not validate their email adress. If it's invalid, you want to avoid sending an email to it to avoid reducing your own email reputation with rejections.

In any case, it's less an 'inactivity' issue, and more 'you did not validate your mail'.

PR #589, opened 2025-03-18
Add checks when removing user from a project's linked group

### LINKED ISSUE #504: Remove from group, then project => error
- Removed myself from a group
- Tried to remove myself from the corresponding project
- Got a " Remove from Project Failed " error
- If I reload the page, I'm removed from the project

rsiminel:
We do not remove the user from the group linked to the project if the user in not in the group anymore.
We do not remove the user from the group linked to the project if it is the user's main group.
- If the user is an admin, we send "Cannot remove primary group xxx. Please adjust the group settings first."
- If the user is not an admin, we send "Cannot change the primary group of xxx. Please contact an admin at xxx@xxx."

closes #504

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Add email sent to a project owner when their project is activated
 * Separate global overview from individual group view in admin group management
 * Allow accents in user's first and last names
+* Add checks when removing user from a project's linked group
 
 ## 1.4.31 (2024-09-27)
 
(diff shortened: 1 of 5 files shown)


mboudet:
Hmm, I'm a bit confused. When you said `If the user is an admin`, are you talking about the user being removed, or the user doing the action?

AFAIK, There are three users who can do this action: Admins, project owners, and users in the projets (who can remove themselves).

I believe the 'Cannot remove primary group xxx. Please adjust the group settings first' message should be sent when an admin is doing the action? But it's currently sent when an admin is removed from the projet.

(You might need to pass an 'is_admin' variable to the remove_user_from_project function to do so)

mboudet (approved):

PR #600, opened 2025-04-30
Tp account deletion double plugin run fix

### LINKED ISSUE #575: Plugins are launched twice on TP account deletion
Here: https://github.com/genouest/genouestaccountmanager/blob/fa802ec249a82c540eb8845d96ec3148f5f01f15/core/tps.service.js#L214

(the delete_user() part also trigger the plugins).

Does not seems to occurs every time, maybe a race condition?

rsiminel:
The `run_plugins` function is run twice when deleting an account, once on the condition that the user is active in the `usrsrv.delete_user` function, and once just after `usrsrv.delete_user` is called.

closes #575


mboudet:
Can you add to the changelog?

mboudet (approved):

Issue #515, opened 2024-09-27
Create a flag for user accounts "extend_not_allowed" 


roadn:
This would enable us to follow accounts from private company whi have access to the cluster through a training for example but that we do not wish to extend beyond the training period

roadn:
ideally we would  be able to automatically delete these user accounts on a set date

roadn:
or make it impossible for the users with this flag to renew their accounts

mboudet:
Seems fine to me to add a 'disabled_self_extend' flag on some accounts. The renew button would be removed.
Do we want to send an email warning of expiration (without the link, just saying "contact an admin") or not?

(I'd say no)

roadn:
I'd also say no.

mitaahs:
Resolved by https://github.com/genouest/genouestaccountmanager/pull/650.
After discussing it, we decided to always send the user an email explaining that the account will be closed and to contact an administrator if necessary.

PR #650, opened 2026-06-12
feat(self-extension rights management): implement user self-extension rights system

mitaahs:
- Add a default value in configuration files to support the new rights management system
- Add a disable_extend field for new users
- Add enforcement mechanisms to validate self-extension rights during extension requests, ensuring only authorized users can extend their accounts
- Add a button on the user page for administrators to toggle self-extension rights
- Exclude the extend button for users without self-extension rights
- Add a dedicated email template for users without self-extend rights approaching account expiration
- Add verification logic to send the appropriate expiration email based on user rights status

Resolves : #618 and #515.


manager2/src/app/user/user.component.html
```diff
@@ -20,7 +20,9 @@ <h2>{{ user.uid }}</h2>
     </div>
     <div class="col-sm-2">
       <button *ngIf="session_user.is_admin && (user.status == STATUS_PENDING_APPROVAL || user.status == STATUS_PENDING_EMAIL)" class="p-button p-button-sm p-button-success" (click)="activate()">Activate</button>
-      <button *ngIf="user.status == STATUS_ACTIVE" class="p-button p-button-sm p-button-success" (click)="extend()">Extend</button>
+      <button *ngIf="session_user.is_admin || (user.status == STATUS_ACTIVE && user.disable_extend === false)"  class="p-button p-button-sm p-button-success" (click)="extend()">Extend</button>
```
mboudet: The 'ng-if' logic is a bit different. Even for admin, the button should only pop if the status is active (we have the renew button otherwise)

So user.status == STATUS_ACTIVE && (session_user.is_admin || user.disable_extend === false)

manager2/src/app/user/user.service.ts
```diff
@@ -57,7 +58,7 @@ export class User {
         projects: string[] = null, new_project: any = null, tags: any = null,
         reg_key: number = 0, api_key: number = 0, ssh: string = '', u2f: any = null, otp: any = null,
         history: any[] = [], extra_info: any[] = [], status: string = '', temp: any = null, loginShell: string = '',
-        maingroup: string = "", uidnumber: number = 0
+        maingroup: string = "", uidnumber: number = 0, disable_extend: boolean = true
```
mboudet: Might want to default to false (current state)

manager2/src/app/user/user.service.ts
```diff
@@ -96,7 +97,7 @@ export class UserService {
             resp.projects || null, resp.newproject || null, resp.tags || null,
             resp.regkey || 0, resp.apikey || 0, resp.ssh || '', resp.u2f || null, resp.otp || null,
             resp.history || [], resp.extra_info || [], resp.status || '', resp.temp || null, resp.loginShell || '',
-            resp.maingroup || '', resp.uidnumber || 0
+            resp.maingroup || '', resp.uidnumber || 0, resp.disable_extend
```
mboudet: I think it's missing a || false, to provide a default value

manager2/src/app/user/user.service.ts
```diff
@@ -468,4 +469,22 @@ export class UserService {
...
+    changeUserExtendRight(userId: string, checked: boolean) {
+        let httpOptions = {
+            body: {
+                disable_extend: checked
+    }
+        }
+        return this.http.post(
+            '/user/' + userId + '/extend-rights', httpOptions.body)
+        
+    }
```
mboudet: Two things:

* I'm not sure why you need to get the api call to get the state. It should have been passed along the user (or gotten a default value)
* I'm not sure we need a dedicated api endpoint to change the value. Maybe we can just add the checkbox at the bottom of the user information (like internal user, etc), and just send it when 'updating' the user like usual?

manager2/src/app/user/user.component.ts
```diff
@@ -126,6 +126,7 @@ export class UserComponent implements OnInit {
     timeoutId: any;
     note: string;
     err_note: string;
+    checked: boolean = true;
```
mboudet: This is basically the value of the user.disable_extend, right? Might not need a dedicated variable

manager2/src/app/user/user.component.ts
```diff
@@ -272,6 +273,15 @@ export class UserComponent implements OnInit {
...
             );
+        
+            this.userService.getUserExtendRight(params['id']).subscribe(
+                (disable_extend: boolean) => {
+                    this.checked = disable_extend;
+                },
+                (err) => {
+                    console.error('Error loading user self-extend right', err);
+                }
+            );
```
mboudet: See comment in user.service

routes/users.js
```diff
@@ -776,7 +776,8 @@ router.post('/user/:id', async function (req, res) {
         duration: req.body.duration,
         history: [{ action: 'register', date: new Date().getTime() }],
         extra_info: req.body.extra_info || [],
-        registration: new Date().getTime()
+        registration: new Date().getTime(),
+        disable_extend: CONFIG.general.default_disable_extend
```
mboudet: Maybe add a || false, to provide a value if the config is not set

manager2/src/app/user/user.service.ts
mitaahs: My line of thinking was as follows: if an error occurs when loading and this default value is used, would I rather have the admin think the user is blocked when they aren't, or vice versa? I figured that there wouldn't be many blocked users, so a user who is blocked when they shouldn't be would seem more unusual to the admin. Which would lead to the admin trying to refresh the page. But I see why this can cause a problem in the case where a user is shown blocked by error.
But I see that this could be a problem if the admin sees a user who is not already blocked has blocked by mistake, especially if they had just come to block that user.

manager2/src/app/user/user.service.ts
mitaahs: - For the api call to get the state : I might be wrong, but I did this because we're using a JWT. I think it can last up to two days. So if we retrieve the information stored on the user’s end without making another call, the information might be out of date. If, in the meantime, the administrator has changed the user’s permissions, the user won’t know about it and will still see the “Extend” button. The button won’t work, which could cause a problem.
But maybe I missed something looking at the console, and the user actually gets that information every time he loads the page?

- For the api to change the value : Yes, I hadn't thought of doing it that way but it's the most obvious. I will do it that way. I had assumed that I needed to create a button next to the "extend" button, and I built everything from there.

manager2/src/app/user/user.service.ts
mboudet: I feel like if an error occurs, the whole user will be default values, so there is not much risk an admin would act on it anyway.

manager2/src/app/user/user.service.ts
mboudet: Nope, I don't think we actually use the info from the jwt token. When loading the user page, there is a call to the api to get the user info : https://github.com/genouest/genouestaccountmanager/blob/master/manager2/src/app/user/user.component.ts#L292

(Which call getUser). So, there should be not issue of desync, you can just rely on what is passed from the API

manager2/src/app/user/user.component.ts
mitaahs: This is necessary because I use a switch button. So this value is bind to the button and is used to keep track of the button state. But if I change the code with the new method, I will try to remove it.

mboudet (approved):

PR #674, opened 2026-08-03
 build(update images): update my and mycron to current debian and node versions

mitaahs:
- Allows to fix numerous security issues
- Allows to rely upon an official node image with known build steps
- Allows to fix the test env that was not usable locally
- The new ldap-utils versions disable the usage of ldapadd -h and replace it by ldapadd -H. This PR adapt the templates to use the new argument.
- Facilitates local usage of the repo for dev and test by allowing the usage of newer ldap-utils versions.
- This PR standardize the name of the my and mycron docker images


cron/Dockerfile-cron
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/genouest/my
+FROM genouest/my
```
mboudet:
```suggestion
FROM quay.io/genouest/my
```
it should still be a quay.io image. We use this Dockerfile when deploying to prod, in addition to the test

.github/workflows/test.yml
```diff
@@ -24,8 +24,8 @@ jobs:
         npm install
         eslint routes
         eslint core
-        sudo docker build -t osallou/my .
-        cd $GITHUB_WORKSPACE/cron && sudo docker build -f Dockerfile-cron -t osallou/mycron .
+        sudo docker build -t genouest/my .
```
mboudet: Same as the other comment: the tag should be quay.io/genouest/my , to match the Dockerfile-cron
```suggestion
        sudo docker build -t quay.io/genouest/my .
```

tests/docker-compose.yml
```diff
@@ -80,7 +80,7 @@ services:
             - LDAP_SERVER_HOST=openldap
 
     my-app:
-        image: osallou/my
+        image: genouest/my
```
mboudet:
```suggestion
        image: quay.io/genouest/my
```

.travis.yml
mboudet, line 1: I don't think we use travis anymore? Maybe we should just remove the file

PR #677, opened 2026-08-06
build(test env launcher): add script to automate test environment usage

mitaahs:
- Add script to easily start and stop a fully working test environment with two methods.
- quick method : Automatize the test method that was use previously. Speed up development but not reproducible.
- full method : Uses the GitHub workflow file. This allows to reproduce exactly what will happen on the GitHub CI pipeline after the push.
- Add Dockerfile necessary to build the image used as the runner for the full method.
- Change README.md with usage instruction for the script.
- Update .gitignore to take into account act installation when using the full method.


tests/test_env_launcher.sh
```diff
@@ -0,0 +1,136 @@
...
+        exit 1
+    fi
+}
+
+if ! [ "$(basename "$PWD")" = "tests" ]; then
+    echo "Please launch this script from the genouestaccountmanager/tests folder."
+	exit 1
+fi
+
+check_distro_packages_installation "docker-cli"
```
mboudet: If this is the only package required (and we just need the 'docker' command, I think?), maybe we can just check if 'docker' is in the path, instead of having a check on the package itself?

(But to be honest, i feel like we can skip the checks. You'd need to check for the sudo package too)

mitaahs: Yes, maybe I have been a bit overzealous! I’ll delete that and amend the commit.

.gitignore
```diff
@@ -11,3 +11,4 @@ manager2/node_modules
 manager2/dist
 templates/default/user/*
 dev_quickstart/env.sh
+/tests/local_testing/act
```
mboudet: Not sure about the first '/' here?

mitaahs: Thanks, it's modified :)

mboudet (approved):

PR #678, opened 2026-08-06
fix(checkmark not displaying) : checkmarks now displaying user state

mitaahs:
- All the checkmarks are now displaying the real user state when an admin looks at a user page.


mboudet:
I think the 'extra' fields are also broken in prod (but I cannot check now)

mitaahs:
The issue with the extra fields is fixed with this new commit.

